PR #1, opened 2019-07-16
Add threat model section

samngmco:
@tomtau see the new Threat Model section


tomtau (approved):
looks ok for the initial version

PR #5, opened 2019-07-26
Problem: no "getting started" documentation extended context (cro-223)

tomtau:
Solution: ported additional explanation from the internal slides

### DIFF
--- a/docs/client-flow.md
+++ b/docs/client-flow.md
@@ -1,0 +1,32 @@
+# Client Flow
+
+## Block-level filtering
+The full node "tags" each block with a probabilistic filter (tagging is done using the [Events](https://tendermint.com/docs/spec/abci/abci.html#events) in Tendermint).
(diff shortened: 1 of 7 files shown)


docs/client-flow.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Client knows transaction
+
+Each block header includes an application hash ("APP_HASH", see [consensus](consensus.md) for details). As a part of it is a root of a Merkle tree of valid transactions, a client can check whether its known transaction was included in a block:
+
+1. get the block's application hash / header information
+
+2. compute the transaction ID from transaction data (see [transaction](transaction.md)).
+
+3. check a Merkle inclusion proof where the transaction ID is one of the leaves, and a part of the application hash is the root
```
devashishdxt, line 23: How does the client gets inclusion proof for a transaction ID?

devashishdxt (approved):

docs/client-flow.md
tomtau: can query with "proof=true" and get it in the response
OR can construct the whole valid tree from transaction IDs

PR #7, opened 2019-08-06
Problem: broken links in the technical documentation

brianMCO:
Solution: Update all links by using relative urls


tomtau (approved):
looks good, thanks a lot!

PR #16, opened 2019-09-13
Problem: (CRO-432) Missing wallet setup and usage instructions

calvinaco:
Solution: Added instructions for @lezzokafka for further polishing

- ClientCLI with command instructions
- ClientRPC with images and steps


lezzokafka (approved):

tomtau (changes requested):
I guess one of those should be linked from https://github.com/crypto-com/chain-docs/blob/master/docs/.vuepress/config.js
so that it's accessible / rendered in the web docs

PR #23, opened 2019-09-19
Problem: (CRO-432) Missing ClientRPC documentation for wallet

calvinaco:
Solution: Added the ClientRPC documentation


tomtau (approved):
👍 documenting client-cli environment variables was much needed

PR #31, opened 2019-11-04
Problem: missing information about validator set changes

tomtau:
Solution: updated transaction types + added a brief note in staking

### DIFF
--- a/docs/getting-started/staking.md
+++ b/docs/getting-started/staking.md
@@ -25,13 +25,23 @@ Council Node is a data structure that holds state about a node responsible for t
 
 ```
 pub struct CouncilNode {
+    pub name: String, // human-readable name
+    pub security_contact: Option<String>, // optional email for security contact
     pub staking_address: RedeemAddress, // account with the required staked amount
     pub consensus_pubkey: crypto.PubKey, // Tendermint consensus validator-associated public key: note that Tendermint supports different signature schemes
     pub council_pubkey: PublicKey, // key for council node-related transaction types
     pub service_whitelist_pubkey: PublicKey, // key for service node whitelist-related transaction types
-    pub nonce: usize, // update counter
+    pub nonce: usize, // update counter?
 }
 ```
+### Joining network
+Anyone can submit a `CreateCouncilNodeTx` that will be valid as long as:
+
+* the associated staking account has bonded amount >= `COUNCIL_NODE_MIN_STAKE` and is not punished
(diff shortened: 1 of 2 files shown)


lezzokafka (approved):

docs/getting-started/staking.md
```diff
@@ -25,13 +25,23 @@ Council Node is a data structure that holds state about a node responsible for t
...
     pub service_whitelist_pubkey: PublicKey, // key for service node whitelist-related transaction types
-    pub nonce: usize, // update counter
+    pub nonce: usize, // update counter?
 }
 ```
+### Joining network
+Anyone can submit a `CreateCouncilNodeTx` that will be valid as long as:
+
+* the associated staking account has bonded amount >= `COUNCIL_NODE_MIN_STAKE` and is not punished
+* there's no other validator with the same `staking_address` or the `consensus_pubkey`
```
samngmco, line 41: `name` should be unique as well

samngmco (commented):
`name` should be unique as well

docs/getting-started/staking.md
lezzokafka: It seems you can have the same Moniker for being a cosmos validator, the identity of the node ended up determined by its pub_key. 
https://cosmos.bigdipper.live/validator/cosmosvaloper190r3wrl2jexq9hyahl4pd5wz66rrt4wpuavdxa
https://cosmos.bigdipper.live/validator/cosmosvaloper12etdez5nurn9avdjrnuezzttck59s0spsx4xea

But I agree with that, to avoid confusion.

tomtau: @samngmco @lezzokafka any particular concern? Even if it's unique, one can similarly looking names?

docs/getting-started/transaction.md
```diff
@@ -32,7 +32,7 @@ The initial prototype uses a linear fee system, see [staking](./staking) for det
 
 ## Transaction Types
 
-### Basic Types (unencrypted / non-shielded):
+### Basic Types (plain version):
```
landanhu, line 35: does `unencrypted / non-shielded` === `plain`?

tomtau: yes

docs/getting-started/staking.md
lezzokafka: @tomtau  Nah, mostly is for delegators in their context.  
You can still get some very similar moniker with an extra emoji 🤔🤔🤔etc.

PR #34, opened 2019-12-02
Add incentive design docs

yihuang:


tomtau (changes requested):
this doc: https://crypto-com.github.io/getting-started/staking.html#rewards contains old sketch, so probably modify it there, or break down the outdated staking document into several pages

yihuang:
> this doc: https://crypto-com.github.io/getting-started/staking.html#rewards contains old sketch, so probably modify it there, or break down the outdated staking document into several pages

Done, I've updated the rewards section in staking docs.

lezzokafka (approved):
Looks good to me, thanks Yi :)
I will further polish it, no worries.

docs/getting-started/staking.md
```diff
@@ -88,55 +64,54 @@ For Advanced TX types (council node and service node state metadata management),
 
 ## Rewards
 
-The rewards pool is a data structure that stores the information about remaining funds:
+To incentivise validators to run the network, rewards are accumulated and distributed to the validators. There are three sources for the rewards:
+
+- Monetary expansion with fixed max supply
+- Transaction Fees
+- Slashings of byzantine nodes
```
tomtau:
```suggestion
- Slashing of byzantine and non-live nodes
```

docs/getting-started/staking.md
```diff
@@ -88,55 +64,54 @@ For Advanced TX types (council node and service node state metadata management),
...
+
+The `RewardsPool` data structure stores all the information about remaining funds and distribution states:
 
 ```
 pub struct RewardsPool {
-    pub nonce: usize, // update counter
-    pub total_remaining: Coin, // total available amount
-    pub last_update: Timespec, // when was the pool last updated
+    pub remaining: Coin, // total available amount
```
tomtau:
```suggestion
    pub period_bonus: Coin, // rewards accumulated from fees and slashing
```

docs/getting-started/staking.md
yihuang: Done

docs/getting-started/staking.md
```diff
@@ -88,55 +64,54 @@ For Advanced TX types (council node and service node state metadata management),
...
+tau(0) = 10000000
+```
 
-- council nodes: proportional to the signed block count (node block signatures / all block validators signatures)
+All the new coins minted in this process is recorded in the `minted` field, and the fixed max supply is checked, after max supply get reached, no new coins will get minted.
 
-:::tip NOTE
-The draft technical whitepaper says there will be rewards for acquirer nodes. These terms may be revised in later drafts.
-:::
+The rewards validator receives goes directly to the unbonded balance of their staking account.
```
tomtau: going to `bonded` balance may be preferable, as one will get a compounding effect (the higher voting power, the more often selected for proposing blocks) + a flexibility of management when to unbond and withdraw rewards from multiple periods.
Or any reason why in unbonded @yihuang ?

tomtau (changes requested):
looks ok, but one small clarification of where the rewards go before merging it and @lezzokafka editing further

docs/getting-started/staking.md
yihuang: I just think unbonded would be simpler and a little bit less risky then bonded, because the later will change voting power directly. Go to `bonded` is indeed more friendly to user.

yihuang: I think we can change to `bonded`, makes more sense that way.

tomtau: Why is unbonded less risky? It seems more risky, because for bonded, one needs to wait a unbonding period before being able to withdraw during which period byzantine faults may be detected -- e.g. one may accumulate rewards in unbonded until there's a lot, launch an attack and withdraw immediately (so only the bonded amount will be slashed rather than the full amount)

docs/getting-started/staking.md
```diff
@@ -88,55 +64,54 @@ For Advanced TX types (council node and service node state metadata management),
...
+S: total stakings at current block when reward distribution happens
 ```
 
-`compute_share` determines the amount for each claiming node; the initial prototype will compute it:
+The parameter `tau` will decay each time rewards get distributed:
+
+```
+tau(n) = tau(n-1) * 0.95
+tau(0) = 10000000
+```
```
tomtau: is this one correct @lezzokafka ? I thought the decay parameter update may be less frequent than reward distribution period?

docs/getting-started/staking.md
tomtau: The only thing about changing the voting power is that if the reward distribution period is too frequent (e.g. after every block), it'll be inefficient, as one will keep generating validatorupdates

docs/getting-started/staking.md
yihuang: It seems the `0.95` was for a year period, it needs to be adjusted to a day period, like `0.99986`, but we can't use `Millis` anymore, maybe use some third-party fixed-point number library.

lezzokafka: @tomtau In this example, *0.95 is the drops per annum.
At the moment it is updated every reward epoch ( sth like 0.95^(1/365) ~0.9986 each day) to provide a "smoother" drops.

If It sounds too frequent, we can make it configurable of how often this decay parameter changes (every $n$ epoch etc.) .

lezzokafka (approved):

PR #41, opened 2019-12-31
(CRO-633) Problem: Chain-docs don't cover voting power

lezzokafka:
Solution: Added _Voting power and proposer selection_  section.

### DIFF
--- a/docs/getting-started/staking.md
+++ b/docs/getting-started/staking.md
@@ -42,9 +42,21 @@ Anyone who wishes to become a council node can submit a NodeJoinTx; this transac
 ####  Voting power and proposer selection
 At the beginning of each round, a council node will be chosen deterministically to be the block proposer.
 
-The chance of being a proposer is directly proportional to their *voting power* at that time, which, in general, is equal to the `bonded` amount in the associated staking address of the council node.
+The chance of being a proposer is directly proportional to their *voting power* at that time, which, in general, is equal to the `bonded` amount (rounded to the whole unit) in the associated staking address of the council node.
 
-The top `MAX_VALIDATORS` (ordered by the voting power) would put to the active validator set in `END_BLOCK`.
+The top `MAX_VALIDATORS` with the most `bonded` amount would put to the *active* validator set in `END_BLOCK`. If a validator's `bonded` amount is below the top `MAX_VALIDATORS`, It will be considered as *non-active* and would not be able to take part in the consensus. For example, 
+
+- If the number of *active* validators < `MAX_VALIDATORS`:
+
+    |    |  `bonded` amount < `COUNCIL_NODE_MIN_STAKE` |  `bonded` amount => `COUNCIL_NODE_MIN_STAKE` | 
+    |----|---|---|
+    | Validator's voting power |  Set to 0 |  Set to the `bonded` amount| 
+
+- On the other hand, If the number of *active* validators = `MAX_VALIDATORS`: 
+
+    |    |  `bonded` amount is below the top `MAX_VALIDATORS` |  Otherwise | 
+    |----|---|---|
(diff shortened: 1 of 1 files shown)


docs/getting-started/staking.md
```diff
@@ -36,9 +36,14 @@ pub struct CouncilNode {
...
 Anyone who wishes to become a council node can submit a NodeJoinTx; this transaction is considered to be valid as long as:
 
-1. The associated staking account has bonded amount >= `COUNCIL_NODE_MIN_STAKE` and is not punished
+1. The associated staking account has `bonded` amount >= `COUNCIL_NODE_MIN_STAKE` and is not [punished](#punishments);
 1. There is no other validator with the same `staking_address` or the `consensus_pubkey`
 
+####  Voting power and proposer selection
+At the beginning of each round, a council node will be chosen deterministically to be the block proposer.
+
+The chance of being a proposer is directly proportional to their *voting power* at that time, which, in general, is equal to the `bonded` amount in the associated staking address of the council node.
```
tomtau: equal -- perhaps `proportional`? just to clarify that the amounts are in full units / coins, not base units / carsons

docs/getting-started/staking.md
```diff
@@ -273,8 +278,8 @@ Besides committing all the relevant changes and computing the resulting `APP_HAS
 
  For example, if the changes are relevant to the `bonded` amount of the council node’s staking address and the validator is not jailed: 
 
-- If the `bonded` amount changes and < `COUNCIL_NODE_MIN_STAKE`, then the validator’s power should be set to 0;
-- If the `bonded` amount changes and >= `COUNCIL_NODE_MIN_STAKE`, then the validator’s power should be set to that amount.
+- If the `bonded` amount changes and < `COUNCIL_NODE_MIN_STAKE`, then the validator’s voting power should be set to 0;
+- If the `bonded` amount changes and >= `COUNCIL_NODE_MIN_STAKE`, then the validator’s voting power should be set to that amount.
```
tomtau: this is a bit imprecise with regards to that `max_validators` network argument -- it's either minimal stake network parameter or the lowest stake if there are max validators

tomtau: I don't know if that's mentioned elsewhere in the text

docs/getting-started/staking.md
tomtau: or that "amounts" mean full coin amounts could be stated somewhere in the overall text

tomtau (changes requested):
looks ok, just two small suggestions

docs/getting-started/staking.md
lezzokafka: Yes, I thought about this earlier as well, in the case of the node is not being punished,  the voting power will be _"Set to the bonded amount"_ , see [END/COMMIT_BLOCK_STATE_UPDATE](https://crypto-com.github.io/getting-started/staking.html#%E2%80%9Cglobal-state%E2%80%9D-app-hash). 
May i ask what's the actual implementation that gives us the voting power?  @devashishdxt

lezzokafka: btw if it is being scaled up, as suggested in the tendermint doc: 
> Voting power uses an int64 but must be positive, thus the range is: 0 through 9223372036854775807. Because of how the current proposer selection algorithm works, we do not recommend having voting powers greater than 10^12 (ie. 1 trillion).

tomtau: @lezzokafka https://github.com/crypto-com/chain/blob/master/chain-core/src/init/coin.rs#L268

docs/getting-started/staking.md
lezzokafka: Cool cool, added the case of maximum validator is reached.

tomtau (approved):

PR #42, opened 2020-01-02
(CRO-625)Problem: no docs for validator node join process

lezzokafka:
Solution: 
 Added
- Node joining Tx;
- Unjailing Tx;
- Simple deposit/withdraw.

for client-cli


tomtau (approved):
this doesn't document the "process" per say, it just documents the command to generate the transaction

Issue #43, opened 2020-01-08
Problem: outdated getting started docs


tomtau:
@foreseaz feel free to submit a PR with thing you recorded

tomtau:
also more context for developers

foreseaz:
@tomtau what updates are needed?

tomtau:
it was re-opened by a mistake, sorry

PR #49, opened 2020-01-13
Problem: outdated getting started docs

foreseaz:
Solutions: update getting started, only include env setup with Azure. Issue [#43]


foreseaz:
The new getting-started page preview: https://crypto-chain-doc.surge.sh/getting-started/#pre-requisites

The logic behind is that `getting started` is only the env setup, and then will move to `run the chain & send 1st tx`. Will update `run chain & send tx` later.

lezzokafka:
Frist PR! @foreseaz

docs/getting-started/README.md
```diff
@@ -3,363 +3,274 @@
...
 
-```bash
-~/bin
-```
+We do recommend you use the cloud vendors listed in the [above reference](https://github.com/ayeks/SGX-hardware#cloud-vendors) to run the node.
+The following document will walk you through running a full node on [Azure Confidential Compute VM](https://azuremarketplace.microsoft.com/en-us/marketplace/apps/microsoft-azure-compute.confidentialcompute?tab=Overview), which comes with the handy `Open Enclave SDK`. Other cloud computing instances such as `IBM Cloud Bare Metal Instances` and `Alibaba Cloud Bare Metal Instance` could surely run the node, however, the Azure Confidential Compute VM is the most affordable one for now.
+
+## The big picture
+
+If you are the first time to setup a node of Crypto Chain, a warm reminder is that
```
lezzokafka:
```suggestion
If you are the first time to setup a node of Crypto.com Chain, a warm reminder is that
```

docs/getting-started/README.md
```diff
@@ -3,363 +3,274 @@
...
+it may take you one hour or two to fully setup and run the Chain. But, `DON'T PANIC` we will
+guide you through it.
+
+Before diving into details and ~~possibly~~ dependency hell, we would like to introduce you the big picture of Crypto Chain's main components as following:
```
lezzokafka:
``` suggestion 
Before diving into details and ~~possibly~~ dependency hell, we would like to introduce you the big picture of Crypto.com Chain's main components as following:
```

docs/getting-started/README.md
```diff
@@ -3,363 +3,274 @@
...
-## Compile enclave
+![](./assets/azure_setup_2.png)
 
-1. git clone [chain-tx-enclave](https://github.com/crypto-com/chain-tx-enclave), activate docker
+Then choose your desirable VM `location`, `size`, `storage` and `network configs`, or you can leave
+ them as default.
 
+## VM environment setup
+
+SSH to the Azure VM, and start the environment setup for Crypto Chain.
```
lezzokafka:
``` suggestion
SSH to the Azure VM, and start the environment setup for Crypto.com Chain.
```

docs/getting-started/README.md
```diff
@@ -3,363 +3,274 @@
...
 
-3. Install sdk. You can install intel-sgx-sdk, and compile the application.
+### Parameters Explaination
+
+- `SGX_MODE`:
+  - `SW` for Software Simulation mode
+  - `HW` for Hardware mode **(✓ for Azure)**
+- `NETWORK_HEX_ID`: Network HEX Id of Tendermint **(`AB` for testing)**
+- `APP_PORT`: Listening Port inside the Docker instance **(Default: 25933)**
+
```
lezzokafka: Can link to [Chain ID and Network ID](https://crypto-com.github.io/getting-started/chain-id-and-network-id.html#chain-id-and-network-id)

lezzokafka:
Maybe also hide the SPID in the picture?  
`docs/getting-started/assets/intel_sub.png`

docs/getting-started/README.md
foreseaz: Thanks for the review!

docs/getting-started/README.md
```diff
@@ -3,363 +3,274 @@
...
 
-By following this tutorial, you can compile and run Crypto.com chain from the scratch. 
+By following this tutorial, you can compile and run Crypto.com chain from scratch.
 
 ## Pre-requisites
 
-Prepare ubuntu 18.0.x and Intel CPU
+Because we utilize the technology of `Intel® Software Guard Extensions (Intel® SGX)`
+for security concerns, the pre-requisites are a little more strict than the other
+chains' setup. A special type of hardware is needed and the reference of [SGX-hardware](https://github.com/ayeks/SGX-hardware)
```
tomtau: "security concerns" is a bit imprecise -- it's for payment data confidentiality... the relevant doc (transaction-privacy.md / https://crypto-com.github.io/getting-started/transaction-privacy.html#motivation ) could also be linked here

docs/getting-started/README.md
```diff
@@ -3,363 +3,274 @@
 ::: warning Caution
 This page is for the development environment set up only, and it is subject to changes.
 
-For anyone interested in joining the Crypto.com chain testnet, 
+For anyone interested in joining the Crypto.com chain testnet,
 please refer to our [testnet documentation](./thaler-testnet).
 :::
 
-By following this tutorial, you can compile and run Crypto.com chain from the scratch. 
+By following this tutorial, you can compile and run Crypto.com chain from scratch.
```
tomtau: probably add / highlight *latest version* and *for development*... as it's not necessary for other use cases

docs/getting-started/README.md
```diff
@@ -3,363 +3,274 @@
...
+Because we utilize the technology of `Intel® Software Guard Extensions (Intel® SGX)`
+for security concerns, the pre-requisites are a little more strict than the other
+chains' setup. A special type of hardware is needed and the reference of [SGX-hardware](https://github.com/ayeks/SGX-hardware)
+could help you identify if your current hardware supports `Intel® SGX` or not.
 
-```bash
-~/bin
-```
+We do recommend you use the cloud vendors listed in the [above reference](https://github.com/ayeks/SGX-hardware#cloud-vendors) to run the node.
+The following document will walk you through running a full node on [Azure Confidential Compute VM](https://azuremarketplace.microsoft.com/en-us/marketplace/apps/microsoft-azure-compute.confidentialcompute?tab=Overview), which comes with the handy `Open Enclave SDK`. Other cloud computing instances such as `IBM Cloud Bare Metal Instances` and `Alibaba Cloud Bare Metal Instance` could surely run the node, however, the Azure Confidential Compute VM is the most affordable one for now.
```
tomtau: It's a public website, so I'd avoid this sort of public endorsements without previous approvals -- besides, I'm not even sure if Azure is the most affordable; some time ago, the quoted prices of the other vendors were similar.
The only more affordable one was OVHCloud.

As this is for development only, it's also not necessary / recommended to use cloud -- the best is your own machine if possible :) 

So instead of "we do recommend".... I'd rephrase it as something like "If your development machine does not support SGX, we recommend spinning up a cloud instance. In this guide, we will walk through the process of setting it up on Azure."

docs/getting-started/README.md
```diff
@@ -3,363 +3,274 @@
...
+
+## The big picture
+
+If you are the first time to setup a node of Crypto Chain, a warm reminder is that
+it may take you one hour or two to fully setup and run the Chain. But, `DON'T PANIC` we will
```
tomtau: "fully setup a development environment"

tomtau (changes requested):
left a few comments -- I haven't read through it fully yet, but looks good so far

docs/getting-started/README.md
foreseaz: Great! That's exactly the reviews and feedbacks I am looking for. Please provide more technical background, motivations, how we make the decisions, etc.
I will update this part.

docs/getting-started/README.md
foreseaz: I see, I will update it.

docs/getting-started/README.md
foreseaz: We can actually add a `Glossary` page to define all the terms and variables we used.

docs/getting-started/README.md
```diff
@@ -3,363 +3,275 @@
...
+
+## The big picture
+
+If you are the first time to setup a node of Crypto.com Chain, a warm reminder is that
+it may take you one hour or two to fully setup a development environment and run the Chain. But, `DON'T PANIC` we will
+guide you through it.
+
+Before diving into details and ~~possible~~ dependency hell, we would like to introduce you the big picture of Crypto.com Chain's main components as following:
+
+![](./assets/big_pic.png)
```
tomtau, line 28: the picture is ok, but the arrows are a little bit confusing -- for example, chain-abci doesn't connect to tx-query app wrapper, it only advertises the connection string (ip/hostname:port) for the client

docs/getting-started/README.md
```diff
@@ -3,363 +3,275 @@
...
 
-1. git clone [chain-tx-enclave](https://github.com/crypto-com/chain-tx-enclave), activate docker
+Then choose your desirable VM `location`, `size`, `storage` and `network configs`, or you can leave
+ them as default.
 
+## VM environment setup
+
+SSH to the Azure VM, and start the environment setup for Crypto.com Chain.
+
+- Install `Docker`: you can refer to following document [How To Install and Use Docker on Ubuntu 18.04
```
tomtau, line 54: once you have a Docker, the quickest getting started is just `docker pull cryptocom/chain:latest` (this image has Rust at that needed version, libzmq and Intel SDK) and working inside that container -- you just need to expose it the device or the AESM socket -- see https://github.com/apache/incubator-teaclave-sgx-sdk/wiki/Environment-Setup#docker-setup

docs/getting-started/README.md
```diff
@@ -3,363 +3,275 @@
...
 
-For hw
+Now, your `chain-tx-validation` component is LIVE.
 
-```
-docker run -ti --device /dev/isgx -v ~/chain-tx-enclave/:/root/sgx -it chain-tx /bin/bash
 
-root@docker:/# LD_LIBRARY_PATH=/opt/intel/libsgx-enclave-common/aesm /opt/intel/libsgx-enclave-common/aesm/aesm_service &
-```
+## Setup SGX Environment
```
tomtau: if this guide shows how to do with Azure + Docker, then this isn't necessary.

(and the recommended way would be to grab released binaries from https://01.org/intel-software-guard-extensions/downloads instead of building them from scratch)

docs/getting-started/README.md
```diff
@@ -3,363 +3,275 @@
...
-:::tip NOTE
-If you have come across the error message of *"Unable to read tendermint initial config (genesis)"*, It means that your tendermint is not yet initialized, thus the genesis file is not found. 
+:::tip Enhanced Privacy ID (EPID)
+The Intel SGX attestation service is a public web service operated by Intel for client-based privacy focused usages on PCs or workstations. The primary responsibility of the Intel SGX attestation service is to verify attestation evidence submitted by relying parties. The Intel SGX attestation service utilizes Enhanced Privacy ID (EPID) provisioning, in which an Intel processor is given a unique signing key belonging to an EPID group. During attestation, the quote containing the processor’s provisioned EPID signature is validated, establishing that it was signed by a member of a valid EPID group. A commercial use license is required for any SGX application running in production mode accessing the Intel SGX attestation service.
 :::
 
+In short, you can go to [Intel Portal](https://api.portal.trustedservices.intel.com/EPID-attestation)
+to sign up for the ID and KEY. It won't take you more than 5 mins.
```
tomtau: perhaps clarify this is for IAS in staging / development environment

docs/getting-started/README.md
```diff
@@ -3,363 +3,275 @@
...
- ```
-
-10. The tendermint genesis file is now ready
+### Finally run tx-query-app
 
+Go to `tx-query` binary folder and run the program:
 ```
-ls ~/.tendermint/config/genesis.json
+$ cd ~/chain/chain-tx-enclave/tx-query/bin/
+$ RUST_LOG=debug ./tx-query-app  0.0.0.0:3443 tcp://127.0.0.1:25933
```
tomtau: if using a Docker, some care may be needed (for exposing / mapping the ports).

if not using a Docker, I think it's easier to use `ipc://....` zeroMQ socket on a local machine for inter-process communication

docs/getting-started/README.md
```diff
@@ -3,363 +3,275 @@
...
+[build]
+rustflags = ["-Ctarget-feature=+aes,+ssse3"]
 ```
 
-2. go to working folder
+Now, go to `chain` folder and compile:
 ```
-cd ~/bin
+$ cd ~/chain/
+$ cargo build --release
```
tomtau: no need to build with `--release` for development

docs/getting-started/README.md
lezzokafka: Cool cool, I have listed some [network parameters](https://crypto-com.github.io/getting-started/network-parameters.html) here, we can gather some terms and variables to form a list later.

docs/getting-started/README.md
```diff
@@ -3,363 +3,275 @@
...
 ```
 
+## Install Tendermint
+
+Install Tendermint from binary file:
+
 ```
-overmind s
+$ wget https://github.com/tendermint/tendermint/releases/download/v0.32.1/tendermint_v0.32.1_linux_amd64.zip
+$ unzip ./tendermint_v0.32.1_linux_amd64.zip
```
tomtau: I recommend using the latest released version (0.32.9)

tomtau (changes requested):
looks all right, but the environment setup could be simplified -- I only compiled Intel SDK/PSW once when I needed to compile it with extra debugging support due to some problems in AESM I tried to diagnose... but that's more of a system developer concern.

This guide should be a general developer guide, and there's no need for such depth

docs/getting-started/README.md
```diff
@@ -3,363 +3,275 @@
...
 
+In short, you can go to [Intel Portal](https://api.portal.trustedservices.intel.com/EPID-attestation)
+to sign up for the ID and KEY. It won't take you more than 5 mins.
 
-   
-5. You can now enter the [chain ID](./chain-id-and-network-id.md) to specify the chain. Note that, the chain ID must begin with `main`, `test` or `dev`, and the last two digis must be a hex string. For example, we can choose `dev-crypto-42` as our chain-ID: 
-```
-new chain id( main-ab )=dev-crypto-42
-```
+Make sure what your subscription is [DEV Intel® Software Guard Extensions Attestation Service (Unlinkable)](https://api.portal.trustedservices.intel.com/Products/dev-intel-software-guard-extensions-attestation-service-unlinkable). Your `SPID` and `Primary key` will be shown on the portal as below:
```
lezzokafka, line 108: "Login required" for the link

docs/getting-started/README.md
foreseaz: So using this docker, I can omit the part of `## Setup SGX Environment`? Will try a fresh start now.

docs/getting-started/README.md
```diff
@@ -3,363 +3,275 @@
...
 ```
-modify like this
+$ cd ~/linux-sgx/SampleCode/LocalAttestation
+$ make
+$ ./app
 ```
-export PATH="$HOME/.cargo/bin:$PATH"
-export PATH=$HOME/bin:$PATH
+
+You should get follow result:
```
calvinaco:
```suggestion
You should get the following result:
```

docs/getting-started/README.md
```diff
@@ -3,363 +3,275 @@
...
-wallet 2(12500000000) amount=
-wallet 3(0x3a9cbc08991db830020fcb5b27b0ecdbc42449bc) address=
-wallet 3(12500000000) amount=
-wallet 4(0xb5b875ecf31d824ce54426377e1c205035758351) address=
-wallet 4(12500000000) amount=
-wallet 5(0x6a5c5ae69efce789bcbba4b9dc932c2577b111b7) address=
-wallet 5(12500000000) amount=
+export SGX_MODE=HW
+export NETWORK_ID=AB
+export SPID=YOUR_SPID
```
calvinaco: I suggest to give more visual hint of change needed
```suggestion
export SPID={YOUR_SPID}
```

docs/getting-started/README.md
```diff
@@ -3,363 +3,275 @@
...
+export IAS_API_KEY=YOUR_PRIMARY_KEY
```
calvinaco:
```suggestion
export IAS_API_KEY={YOUR_PRIMARY_KEY}
```

docs/getting-started/README.md
```diff
@@ -3,363 +3,275 @@
...
+$ RUST_LOG=debug ./tx-query-app  0.0.0.0:3443 tcp://127.0.0.1:25933
 ```
 
-### Launch chain
-1. Go to working folder and launch `dev-utils` by 
+The output should be as following:
 ```
-cd ~/bin
-./dev-utils run
+[INFO  tx_query_app] [+] Init Query Enclave Successful 2!
```
calvinaco: Just curious, is the "Successful *_2_*!" suppose to be a debug message?

docs/getting-started/README.md
```diff
@@ -3,363 +3,275 @@
...
+Now, go to `chain` folder and compile:
 ```
-cd ~/bin
+$ cd ~/chain/
+$ cargo build --release
 ```
 
-3. locate Procfile
+
+After successfully complied, you will get the following binary files in `./target/release` folder:
```
calvinaco: Following Tomas comment, If you don' use `--release`, update here to `./target/debug` as well.

calvinaco (changes requested):

docs/getting-started/README.md
tomtau: Yes, pretty much -- on a normal machine, you'd just need to do Step 0 (installing the driver): https://github.com/apache/incubator-teaclave-sgx-sdk/wiki/Environment-Setup#intel-sgx-toolchain-setup if on Azure with OpenEnclave ticked, then it's already installed

docs/getting-started/README.md
tomtau: yes perhaps, Linfeng added logging, but it may need some log level tweaking not to make it too chatty

foreseaz:
Thanks for all the reviews and comments! I just followed through Linfeng's makefile flow and it's really well-organized. I will update a new getting started according to the `make` flow. For the current PR, I will try to extract some parts out as in-depth manual building reference.

foreseaz:
Preview at https://crypto-chain-doc.surge.sh/getting-started/  for new make flow

tomtau (approved):
👍the getting started guide is much simpler with that makefile

foreseaz:
Exactly! Thanks @linfeng-crypto 's efforts!

PR #45, opened 2020-01-09
Problem: testnet instructions not update to the latest version

tomtau:
Solution: drafted an updated version

(not to be merged yet -- the TODO holes will be filled in after @devashishdxt 's deployment)

PR #50, opened 2020-01-13
Problem: (CE-57) Documentation not updated for connecting to latest Thaler testnet

devashishdxt:
Solution: Updated documentation with latest testnet details

### DIFF
--- a/docs/getting-started/notes-on-production-deployment.md
+++ b/docs/getting-started/notes-on-production-deployment.md
@@ -1,6 +1,6 @@
 # Notes on Production Deployment
 
-- See [Tendermint notes on running in production](https://docs.tendermint.com/master/tendermint-core/running-in-production.html#database)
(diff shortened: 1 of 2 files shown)


tomtau (changes requested):
the instructions aren't precise (they are different whether one wants to run a validator node or a full node) -- you can take the draft I wrote: https://github.com/crypto-com/chain-docs/pull/45/files and fill in the gaps / do small edits

tomtau:
> the instructions aren't precise (they are different whether one wants to run a validator node or a full node) -- you can take the draft I wrote: https://github.com/crypto-com/chain-docs/pull/45/files and fill in the gaps / do small edits

It may also be good to add:
```
::: warning CAUTION
... (something something) ...
 :::
```

at the top with some text warning that it's not backwards compatible with 0.0 -- so if the person ran the old node, they'd also need to delete the old data etc.

lezzokafka (approved):
Looks good to me 😺

PR #55, opened 2020-01-21
Updated  big_pic.png with default ports labels 

lezzokafka:
Update the `big_pic.png` with default ports labels as follow:

big_pic


devashishdxt (changes requested):
I think `tx-query` should also be connected to `client-rpc` instead of directly connecting with consumer.

lezzokafka:
> I think `tx-query` should also be connected to `client-rpc` instead of directly connecting with consumer.

Sure, let me rearranged it :D

lezzokafka:
big_pic

devashishdxt (approved):

PR #66, opened 2020-01-30
Problem: no specified behaviour for reward/punishment interaction with validator set changes (fixes #58)

### LINKED ISSUE #58: Problem: no specified behaviour for reward/punishment interaction with validator set changes

tomtau:
Solution: sketched out basic rules of how this would happen
Note that the internal refactoring suggested in https://github.com/crypto-com/chain/pull/966
can be helpful in there
(+ note there's still a lot of other unspecified things (not sure if they affect this interaction) and staking docs are outdated in different places)

### DIFF
--- a/docs/getting-started/staking.md
+++ b/docs/getting-started/staking.md
@@ -334,7 +334,7 @@ If the validator wishes to re-join the validator set, they can (unjail if necess
 
 If the NodeJoinTx transaction is valid AND this happens before  `SLASH_MAP_DELETE` AND the consensus pubkey (or associated Tendermint address) from NodeJoinTx transaction is the same, the previous slashing information is preserved (i.e. the "delete schedule" is cancelled).
 
-If the NodeJoinTx transaction is valid AND this happens before next reward distribution AND the consensus pubkey from NodeJoinTx transaction is different, the associated staked state is then rewarded for interactions with both consensus pubkeys.
+If the NodeJoinTx transaction is valid AND this happens before next reward distribution AND the consensus pubkey from NodeJoinTx transaction is different, the associated staked state is then rewarded for interactions with both consensus pubkeys (as reward tracking is per staked state).
 
 ### Validators / Council Nodes
 
--- a/docs/getting-started/staking.md
+++ b/docs/getting-started/staking.md
@@ -334,7 +334,7 @@ If the validator wishes to re-join the validator set, they can (unjail if necess
 
 If the NodeJoinTx transaction is valid AND this happens before  `SLASH_MAP_DELETE` AND the consensus pubkey (or associated Tendermint address) from NodeJoinTx transaction is the same, the previous slashing information is preserved (i.e. the "delete schedule" is cancelled).
 
-If the NodeJoinTx transaction is valid AND this happens before next reward distribution AND the consensus pubkey (or associated Tendermint address) from NodeJoinTx transaction is different, the associated staked state's bonded amount is for both.
+If the NodeJoinTx transaction is valid AND this happens before next reward distribution AND the consensus pubkey from NodeJoinTx transaction is different, the associated staked state is then rewarded for interactions with both consensus pubkeys.
 
 ### Validators / Council Nodes
 
--- a/docs/getting-started/staking.md
+++ b/docs/getting-started/staking.md
@@ -1,18 +1,19 @@
 # Staking
 Crypto.com Chain is based on Tendermint Core's consensus engine, it relies on a set of validators (Council Node) to participate in the proof of stake (PoS) consensus protocol, and they are responsible for committing new blocks in the blockchain. This document contains the specification of the initial staking, slashing, and rewards mechanisms which establish the foundation of the token ecosystem of CRO.
 
-## Account
+## Staked state / Account
 
-Account is a data structure that holds state about staking:
+StakedState is a data structure that holds state about staking:
 
 ```
-pub struct Account {
-    pub nonce: usize, // transaction counter
-    pub bonded: Coin, // bonded staked amount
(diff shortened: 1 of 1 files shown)


docs/getting-started/staking.md
```diff
@@ -221,6 +230,25 @@ slashing_rate = max(liveness_slash_percent, byzantine_slash_percent) * (sqrt(val
               = 0.2
 ```
 
+#### Removed validators / council nodes
+A Validator is removed when its voting power is set to 0.
+This can happen in the following scenarios:
+
+1. The Validator is being punished for infraction(s).
+1. Its operator submitted UnbondStakeTx with sufficient stake to leave the network.
+1. The current validator's set size reached `MAX_VALIDATORS`, there's a new council node joining
```
calvinaco: Do you want to use `-` instead of `1.` here?

calvinaco (approved):

docs/getting-started/staking.md
```diff
@@ -29,16 +30,24 @@ pub struct CouncilNode {
...
+Effective minimum stake (`EFFECTIVE_MIN_STAKE`) is defined as follows at any time:
+
+1. if the number of validators has not reached `MAX_VALIDATORS`, it is `COUNCIL_NODE_MIN_STAKE` (the network parameter)
+1. if the number of validators has reached `MAX_VALIDATORS`, it is equal to the lowest bonded amount of the staked states of the current validators plus 1.0 Coin.
+
 ### Joining the network
 Anyone who wishes to become a council node can submit a NodeJoinTx; this transaction is considered to be valid as long as:
 
-1. The associated staking account has `bonded` amount >= `COUNCIL_NODE_MIN_STAKE` and is not [punished](#punishments);
+1. The associated staking account has `bonded` amount >= `EFFECTIVE_MIN_STAKE` and is not [punished](#punishments);
```
yihuang, line 45: When the number of validators reached `MAX_VALIDATORS`, it would be useful to allow the user to deposit more than `COUNCIL_NODE_MIN_STAKE` but less than `EFFECTIVE_MIN_STAKE` and send join request to become a validator candidate, who will become a validator automatically when the situation changes.

docs/getting-started/staking.md
yihuang: I think there is no problem to keep recording the voting power of all validators with at least `COUNCIL_NODE_MIN_STAKE` bonded coins, treat them like validator candidates, only filter with `MAX_VALIDATORS` in `end_block` when computing `ValidatorUpdate` for tendermint.

docs/getting-started/staking.md
```diff
@@ -221,6 +230,25 @@ slashing_rate = max(liveness_slash_percent, byzantine_slash_percent) * (sqrt(val
...
+A Validator is removed when its voting power is set to 0.
+This can happen in the following scenarios:
+
+1. The Validator is being punished for infraction(s).
+1. Its operator submitted UnbondStakeTx with sufficient stake to leave the network.
+1. The current validator's set size reached `MAX_VALIDATORS`, there's a new council node joining 
+whose associated bonded stake is larger than the one associated to the Validator.
+
+When this happens, the following metadata is cleaned up as follows:
+- its associated block proposer-related data is cleared in the block when this occurs
```
yihuang: Does this mean the validator loses the reward for the whole reward period? If that's the case, a malicious validator can choose to join just before the end of the reward period, and cause other validators to lose the rewards.

lezzokafka: This is somewhat similar to the case that: If a validator opted-out in between, It would losses the reward due to he/she is not an active validator even it has proposed some blocks?

docs/getting-started/staking.md
tomtau: Given the situation of reaching `MAX_VALIDATORS` doesn't seem to be happening in the near future, I'd consider this candidate waitlist to be more of a "nice to have" rather "must have" in the initial specs + implementation?

docs/getting-started/staking.md
tomtau: it's an extra complexity -- https://github.com/crypto-com/chain-docs/pull/66/files#r373284419 It's nice to have, but I'd omit it for now due to tight deadlines

docs/getting-started/staking.md
tomtau: > malicious validator can choose to join just before the end of the reward period, and cause other validators to lose the rewards.

I don't see how this can happen or how is that malicious ? Let's discuss on Slack (there's a channel dedicated to this design)

docs/getting-started/staking.md
lezzokafka: In the thrid case, since we have already introduced the `EFFECTIVE_MIN_STAKE`, would it be easier to say : 
`bonded` amount <=`EFFECTIVE_MIN_STAKE` ?

docs/getting-started/staking.md
lezzokafka: `EFFECTIVE_MIN_STAKE` is, in fact, a stronger condition when compared to `COUNCIL_NODE_MIN_STAKE`. 
I think in most of the case, we can refer to `EFFECTIVE_MIN_STAKE` in the doc once we have clearly defined it.

docs/getting-started/staking.md
tomtau: yes

docs/getting-started/staking.md
tomtau: yes, there's a couple of places where it is talking about it... so if it's defined in one place, it'll be better to use it

lezzokafka (approved):

leejw51crypto (approved):
lgtm

PR #69, opened 2020-01-31
Problem: reward don't have stable p.a. rate

yihuang:
Solution: added a power function to adjust the rate based on `period_seconds / year_seconds`

### DIFF
--- a/docs/getting-started/staking.md
+++ b/docs/getting-started/staking.md
@@ -48,15 +48,15 @@ The top `MAX_VALIDATORS` with the most `bonded` amount would put to the *active*
 
 - If the number of *active* validators < `MAX_VALIDATORS`:
 
-    |    |  `bonded` amount < `COUNCIL_NODE_MIN_STAKE` |  `bonded` amount => `COUNCIL_NODE_MIN_STAKE` | 
+    |    |  `bonded` amount < `COUNCIL_NODE_MIN_STAKE` |  `bonded` amount => `COUNCIL_NODE_MIN_STAKE` |
     |----|---|---|
-    | Validator's voting power |  Set to 0 |  Set to the `bonded` amount| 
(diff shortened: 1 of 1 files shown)


docs/getting-started/staking.md
```diff
@@ -94,16 +94,31 @@ The recording of block proposer is done in `BeginBlock` right before rewards dis
 
 ### Monetary expansion
 
-Monetary expansion is designed to release tokens from the reserve account to the reward pool, while keeping a fixed max total supply. The number of tokens being released at each period is defined as:
+Monetary expansion is designed to release tokens from the reserve account to the reward pool, while keeping a fixed max total supply.
 
+There are several configirations for this:
```
lezzokafka: Can note that these are `rewards_config` in the genesis.

docs/getting-started/staking.md
```diff
@@ -94,16 +94,31 @@ The recording of block proposer is done in `BeginBlock` right before rewards dis
...
+N = S * (pow(1 + R, P / Y) - 1)
+
+N: released coins
 S: total stakings at current block when reward distribution happens
-R_0: upper bound for the reward rate p.a. 
+R0: upper bound for the reward rate p.a.
 ```
 
-Note that this is an exponential decay function, where the index of  controls the “steepness” of the curve. Precisely, this damping factor controls the exponential decay rate of the reward function.
+Note that this is an exponential decay function, where the index of it controls the “steepness” of the curve. Precisely, this damping factor controls the exponential decay rate of the reward function.
```
lezzokafka, line 139: Should we also cover ` "monetary_expansion_decay"` somewhere?

yihuang: Yeah, I added it into the decay formular.

docs/getting-started/staking.md
yihuang: Done

lezzokafka (approved):

Issue #72, opened 2020-02-01
Problem: No clear instructions on checking balance/ staking stake for client-cli.


lezzokafka:
It was covered in "[Send you first Tx](https://crypto-com.github.io/getting-started/send_your_first_transaction.html#send-your-first-transaction)" ( subject to change #56 )

Which should also be clearly covered in [Funds transfer](https://crypto-com.github.io/wallets/client-cli.html#_2-funds-transfer) and [Staking operations](https://crypto-com.github.io/wallets/client-cli.html#_3-staking-operations) under [Client-cli](https://crypto-com.github.io/wallets/client-cli.html#clientcli).

PR #73, opened 2020-02-03
Problem: No clear instructions on checking balance/ staking stake for client-cli. 

lezzokafka:
#72 
Solution: Added instructions for checking balance and staking state  in `wallets/client-cli.md`


tomtau (approved):
👍

PR #75, opened 2020-02-05
Problem: No specification for validator logic.

yihuang:
Solution:
- Write a specification with pseudo-codes(sql+rust).

Notice:
- punishment calculation detail is TBD (or to be mentioned in another spec)
- validator states are stored with in-memory data structures and serialization, `staked_state` is stored with Merkle trie. There is no nothing-at-stake attack because of the `minimal_required_stake` to become a validator? Or merge validator state and staking state and store them all in Merkle trie, only indexes with the in-memory data structure?

### DIFF
--- a/docs/getting-started/validator-spec.md
+++ b/docs/getting-started/validator-spec.md
@@ -62,7 +62,7 @@ create table validator (
 
 -- Currently chosen validator set and voting powers snapshoted at the end of last block
 create table validator_snapshot (
-  staking_address text primary key,
+  validator_address text primary key,
   voting_power int,
 );
 
--- a/docs/getting-started/validator-spec.md
+++ b/docs/getting-started/validator-spec.md
@@ -1,0 +1,316 @@
+## Spec
+
+Specification for validator logic, using simple relational database concepts(table/index/unique constraint) and pseudo-codes(sql+rust).
+
+## Configs
+
+Config items will be referenced in pseudo code directly.
+
+- `max_validators`
+- `minimal_required_staking`
+- `jail_duration`
+- `slash_wait_period`
+- `reward_period`
+- `unbonding_period`
+
+Config constraints:
+
+- `jail_duration >= slash_wait_period`
+
+  To make sure unjail not happen before slash.
+
+- `slash_wait_period >= max_evidence_age`
+
+  To make sure all the evidences of validator are processed before slash.
+
+- `unbonding_period >= max_evidence_age + slash_wait_period`
+
+  To make sure unbonded coins not withdrawn before  evidences detected and slashed.
+
+### Table schema
+
+Table is a set of rows with auto updated primary and secondary indexes.
+
+```sql
+-- account model
+create table staked_state {
+  staking_address text primary key,
+  bonded: int,
+  unbonded: int,
+  unbond_time: timestamp default=0,
+  -- TODO nonce: int,
+}
+
+-- current validator state
+create table validator (
+  staking_address text primary key,
+  bonded int indexed,
+  validator_address text indexed unique,
+  council_node_info json,
+  inactive_time timestamp null,  -- the block time when node becomes inactive, null means active validator candidate
+  voting_power int = bonded / 10000_0000,  -- computed on the fly
+);
+
+-- Currently chosen validator set and voting powers snapshoted at the end of last block
+create table validator_snapshot (
+  staking_address text primary key,
+  voting_power int,
+);
+
+create table reward_stat (
+  staking_address primary key,
+  blocks int,  -- or votes
+);
+
+create table punishment (
+    staking_address primary key,
+    slash_rate decimal,
+    jail_time timestamp (partial indexed when slash_amount is null),  -- the timestamp validator is jailed.
+    slash_reason text,
+    slash_amount int null,  -- amount slashed, null means slash not executed yet.
+);
+
+last_reward_distribution_time: timestamp = 0
+```
+
+### Init
+
+```
+load all tables
+```
+
+### `commit`
+
+```sql
+save all tables
+```
+
+### `end_block`
+
+#### Set `inactive_time` for newly inactive validators
+
+Find the newly inactive validators, mark them by settings `inactive_time`, after this, `inactive_time is null` means active validator candidate.
+
+```sql
+update validator set inactive_time=:block_time
+  where inactive_time is null  -- active currently
+    and ( bonded < minimal_required_staking  -- not enough bonded coins
+       or exists (select 1 from punishment where staking_address=validator.staking_address)  -- or punished
+        );
+```
+
+#### Generate validator updates
+
+```sql
+-- get new validators
+new_validators = select validator_address, voting_power
+  from validator
+  order by bonded desc, staking_address
+  where inactive_time is null  -- active validator candidate
+  limit max_validators
+
+-- get old validators
+old_validators = select * from validator_snapshot;
+
+replace validator_snapshot with new_validators;
+
+validator_updates = diff_validators(
+    old_validators.collect::<HashMap<_, _>>(),
+    new_validators.collect::<HashMap<_, _>>()
+)
+```
+
+```rust
+fn diff_validators(
+    old: &HashMap<ValidatorAddress, VotingPower>,
+    new: &HashMap<ValidatorAddress, VotingPower>,
+) -> Vec<(ValidatorAddress, VotingPower)> {
+    // updates + removes
+    new.iter()
+        .filter_map(|(addr, power)| {
+            if old.get::<str>(addr) != Some(power) {
+                Some((addr.clone(), *power))
+            } else {
+                None
+            }
+        })
+        .chain(old.iter().filter_map(|(addr, _)| {
+            if !new.contains_key(addr) {
+                Some((addr.clone(), 0))
+            } else {
+                None
+            }
+        }))
+        .collect::<Vec<_>>()
+}
+```
+
+#### Cleanup
+
+```sql
+delete from validator
+  where inactive_time is not null  -- inactive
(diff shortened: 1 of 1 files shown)


docs/getting-started/validator-spec.md
```diff
@@ -0,0 +1,200 @@
...
+```sql
+delete from validator
+  where block_time - inactive_time > max_evidence_age
+    and not exists (select 1 from reward_stat where staking_address=validator.staking_address)
+    and not exists (select 1 from punishment where staking_address=validator.staking_address)
+```
+
+### `deliver_tx`
+
+#### Handle deposit/unbond tx
```
tomtau: what about `unbond` and interaction with the minimal required stake network parameter?

yihuang: That is handled in `end_block` when computing validator updates for tendermint. Basically, it re-compute new validator set, and diff with last one, send the diff to tendermint.

yihuang: Oh, sorry, I think I deleted an important line of code, I'll add it back = ="

EDIT: I was confused, the logic of filtering with `minimal_required_staking` happens in "Update inactive_time for newly inactive validators" section, which happens before "Generate validator updates". So validators with not enough bonded coins or get punished, get marked by setting `inactive_time`, after this, `inactive_time is null` means active validator candidates.

yihuang:
https://github.com/yihuang/chain/blob/validator-spec/chain-core/src/state/validator_spec.rs
A rust implementation for this spec.

EDIT: lagged behind the spec now.

docs/getting-started/validator-spec.md
```diff
@@ -0,0 +1,390 @@
...
+}
+
+-- current validator state
+create table validator (
+  staking_address text primary key,
+  bonded int indexed,
+  validator_address text indexed unique,
+  council_node_info json,
+  -- liveness tracker for active validator, null means inactive validator, array of bool can be optimized with bitvec.
+  liveness_tracker [bool; block_signing_window] null,
```
lezzokafka, line 55: Is there a tracker to indicate the block time when this node becomes active? 
It is a part of ` council_node_info`?

docs/getting-started/validator-spec.md
```diff
@@ -0,0 +1,390 @@
...
+
+Specification for validator logic, using simple relational database concepts(table/index/unique constraint) and pseudo-codes(sql+rust).
+
+## Configs
+
+Config items will be referenced in pseudo code directly.
+
+- `max_validators`
+- `minimal_required_staking`
+- `jail_duration`
```
devashishdxt, line 11: Should we add unit of measurement at the end of each config? For example, `jail_duration_secs`, etc.

docs/getting-started/validator-spec.md
```diff
@@ -0,0 +1,390 @@
...
+
+- `jail_duration >= slash_wait_period`
+
+  To make sure unjail not happen before slash.
+
+- `slash_wait_period >= max_evidence_age`
+
+  To make sure all the evidences of validator are processed before slash.
+
+- `unbonding_period >= max_evidence_age + slash_wait_period`
```
tomtau: @devashishdxt I'm not sure about these constraints.
isn't enough `unbonding_period >= jail_duration`, `jail_duration >= max_evidence_age` and `jail_duration >= slash_wait_period`?

The purpose of jail / unbond time should correspond to the "max evidence duration" in Tendermint consensus parameters (as a standard nothing-at-stake mitigation).
The purpose of `slash_wait_period` was, if I'm not mistaken, to have a "grace" period during which only the worst detected infraction is punished.

devashishdxt: `slash_wait_period` should be greater than `max_evidence_age`. I think following three conditions should be enough (`jail_duration >= max_evidence_age` here is implied).
- `unbonding_period >= jail_duration`
- `jail_duration >= slash_wait_period`
- `slash_wait_period >= max_evidence_age`

lezzokafka: Once a validator being jailed, Its voting power should go zero; which means it can not misbehave in the consensus process any more. Then the `slash_wait_period` starts ticking, wait for evidence (if any) and see if there are any even worst misbehaviour.

docs/getting-started/validator-spec.md
```diff
@@ -0,0 +1,390 @@
...
+  -- 3. deposit coins into staking addresses
+  [Handle `deposit` tx](#handle-deposit-tx)
+
+  -- 4. clear reward stat
+  truncate table reward_stat;
+```
+
+## Rust implementation considerations
+
+- Table is implemented with `Slab<Row>`, unique index/primary key is implemented as `BTreeMap<field, usize>` or `HashMap<field, usize>`, non-unique index is implemented as `BTreeMap<field, Vec<usize>>`, indexing structures need to be maintained manually.
```
devashishdxt, line 388: Are we planning to use these generic table data structures (`Slab<Row>`) in code? If yes, why not directly use concrete `struct` for that table?

docs/getting-started/validator-spec.md
yihuang: `slash_wait_period >= max_evidence_age` is to make sure all the faults being detected before execute slashing.
Regarding `unbonding_period >= max_evidence_age + slash_wait_period`, it is unneeded if withdrawing is forbidden after validator gets jailed, I guess this is the current behavior?
And if this is the case, `unbonding_period >= jail_duration` is also unneeded? Only need `unbonding_period >= max_evidence_age`.

lezzokafka: > `slash_wait_period >= max_evidence_age` is to make sure all the faults being detected before execute slashing.

I agree with this ^

tomtau: > is to make sure all the faults being detected before execute slashing.

 "all faults" during which period?

yihuang: > > is to make sure all the faults being detected before execute slashing.
> 
> "all faults" during which period?

The period of `max_evidence_age` before the first fault detected. It is possible for the faults to be detected out of order?

docs/getting-started/validator-spec.md
yihuang: > Should we add unit of measurement at the end of each config? For example, `jail_duration_secs`, etc.

I agree, but currently the spec is more like a prototype, to see if it makes sense. There are even cases it works differently from the current implementation.

docs/getting-started/validator-spec.md
lezzokafka: `slash_wait_period`
                                <----------------------->
                 (           ) |                         | 
                  <--------->  ^                         ^
                  `max_e_age`  Jail                     Slash

lezzokafka: The period is ` max_evidence_age` before the _Jail_  point

docs/getting-started/validator-spec.md
yihuang: > Are we planning to use these generic table data structures (`Slab<Row>`) in code? If yes, why not directly use concrete `struct` for that table?

I try to make the spec as simple as possible, I just happen to find relational DB language seems simplest here.
For a table, there are well-established concepts like index/join/order/filter, which takes extra rust code to implement.
There is a rust translation of this (it's lagged behind the spec itself though) (https://github.com/yihuang/chain/blob/validator-spec/chain-core/src/state/validator_spec.rs).

docs/getting-started/validator-spec.md
lezzokafka: @yihuang Faults could be detected out of order. The actual case should be something like:  
A validator got charged and jailed due to a lower penalty fault (say liveness) and but we will still open to any other evidence for  `slash_wait_period` and see if it also commit double sign fault. 

(the other way round is trivial due to we have already charge it with the higher one)

The logic to me is, change it first, if you see anything, which also "lockdown" the funds. Then we give it plenty of time to see if we need to penalize it even further.

docs/getting-started/validator-spec.md
lezzokafka: Haha, of course. That is chain [#874](https://github.com/crypto-com/chain/issues/874)

docs/getting-started/validator-spec.md
```diff
@@ -0,0 +1,390 @@
...
+  -- the block time when node becomes inactive, null means active validator candidate
+  inactive_time timestamp null,
+
+  -- computed on the fly
+  voting_power int = bonded / 10000_0000,
+);
+
+-- Currently chosen validator set and voting powers snapshoted at the end of last block
+create table validator_snapshot (
+  staking_address text primary key,
```
lezzokafka: Should be some consensus-related metadata like `validator_address`

docs/getting-started/validator-spec.md
lezzokafka: ` jail_duration`
                               <--------------------------------->

              (     ...    )   ║                         ¦       ║

              <----...---->    ^                         ^       ^
                  `max_e_age`  |                         |       |
                             Jailing                   Slash   Able to unjaill      
                                         
                                <----------------------->
                                   `slash_wait_period`

docs/getting-started/validator-spec.md
yihuang: AFAIK, currently no such info.

docs/getting-started/validator-spec.md
yihuang: > Should be some consensus-related metadata like `validator_address`

Fixed, good catch.

devashishdxt (changes requested):
This now diverges a lot from https://github.com/crypto-com/chain-docs/pull/89. Can you either remove punishemts from this or modify them based on the punishement specifications PR?

yihuang:
> This now diverges a lot from #89. Can you either remove punishemts from this or modify them based on the punishement specifications PR?

Maybe we can close this, and re-do it based on new rules.

Issue #80, opened 2020-02-13
Problem: punishment and monetary network parameters are not explained


tomtau:
monetary params -- unbonding period, fee policy, minimal stakes...

if the outcome depends on network parameters, each parameter needs to be justified:
what it does
what are its constraints
what its relation to Tendermint core consensus security assumptions is -- "nothing-at-stake" problem, consensus-critical parameters set in Tendermint (evidence age, etc.), ...
what it should be set to -- if it is not given (like total maximum supply), the recommended value should be given with justification / reference explaining why, e.g. based on mechanism design simulation models

PR #89, opened 2020-02-19
Problem: Punishments are not defined precisely

devashishdxt:
**Solution**: Simplify network parameters and fill gaps in current punishments specifications. Related to #80, #83 and #84. 

**How it addresses crypto-com/chain#1096?**

The problem with current implementation is that it does not take `evidence`'s `block_height` in account when calculating `slash_ratio` (it assumes `block_height = current_block_height` and takes latest `validator_voting_power` in `slash_ratio` calculations. Voting power at `current_block_height` may be different because a validator can `unbond`/`deposit` funds into their account).

This PR gets rid of proportional slashing and immediately slashes an account when they make a byzantine fault.

**How it prevents bugs like crypto-com/chain#1079?**

This PR gets rid of proportional slashing and immediately slashes an account when they make a byzantine fault.

**Note**: I've created a new file `punishments.md` but haven't linked that in `staking.md`. Fixes in `staking.md` will be done in future PRs.

### DIFF
--- a/docs/getting-started/punishments.md
+++ b/docs/getting-started/punishments.md
@@ -100,3 +100,8 @@ A validator should not be slashed more than once within `UNBONDING_PERIOD`. If a
 within that time period, it should only be slashed once (for simplicity, we'll only slash the validator for the first
 evidence that we get from tendermint and ignore other evidences until `UNBONDING_PERIOD`).
 :::
+
+:::tip Implementation note:
+To enforce that we only slash an account only once within `UNBONDING_PERIOD`, we can just check if an account is
+not jailed when we receive evidence of misbehavior from tendermint.
+:::
--- a/docs/getting-started/punishments.md
+++ b/docs/getting-started/punishments.md
@@ -1,0 +1,153 @@
+# Validator Punishments
+
+This part describes functionality that aims to dis-incentivize network-observable actions, such as faulty validations,
+of participants with values at stake by penalizing/slashing and jailing them. The penalties may include losing some
+amount of their stake (surrendered to the rewards pool), losing their ability to perform the network functionality for a
+period of time, collect rewards etc.
+
+## Network Parameters
+
+Below are all the network parameters used to configure the behavior of validator punishments. Details of all these
+parameters and their effect on behavior of validator punishments is discussed later in this document.
+
+1. `UNBONDING_PERIOD`: Unbonding period will be used as jailing period (time for which an account is jailed after it
+   gets punished) and also as slashing period (time to wait before slashing funds from an account). This should be
+   greater than or equal to `MAX_EVIDENCE_AGE` in tendermint.
+1. `BLOCK_SIGNING_WINDOW`: Number of blocks for which the moving average is calculated for uptime tracking.
+1. `MISSED_BLOCK_THRESHOLD`: Maximum number of blocks with faulty/missed validations allowed for an account in last
+   `BLOCK_SIGNING_WINDOW` blocks before it gets jailed.
+1. `LIVENESS_SLASH_PERCENT`: Percentage of funds (bonded + unbonded) slashed when validator is not live.
+1. `BYZANTINE_SLASH_PERCENT`: Percentage of funds (bonded + unbonded) slashed when validator makes a byzantine fault.
+
+:::tip Important:
+During slashing, funds are slashed from both, bonded and unbonded, amounts.
+:::
+
+## Overview
+
+Punishments for a validator are triggered when they either make a *byzantine fault* or become *non-live*: 
+
+- Liveness Faults (Low availability)
+
+    A validator is said to be **non-live** when they fail to sign at least `MISSED_BLOCK_THRESHOLD` blocks in
+    last `BLOCK_SIGNING_WINDOW` blocks successfully. `BLOCK_SIGNING_WINDOW` and `MISSED_BLOCK_THRESHOLD` are network
+    parameters and can be configured during genesis (currently, changing these network parameters at runtime is not
+    supported). Tendermint passes signing information to ABCI application as `last_commit_info` in `BeginBlock` request.
+
+:::tip Example:
+For example, if `BLOCK_SIGNING_WINDOW` is `100` blocks and `MISSED_BLOCK_THRESHOLD` is `50` blocks, a validator will
+be marked as **non-live** if they fail to successfully sign at least `50` blocks in last `100` blocks.
+:::
+
+- Byzantine Faults (Double signing)
+
+    A validator is said to make a byzantine fault when they sign conflicting transactions at the same height and round.
+    Tendermint has mechanisms to publish evidence of validators that signed conflicting votes (it passes this 
+    information to ABCI application in `BeginBlock` request), so they can be punished by the application.
+
+:::tip Implementation note:
+Tendermint passes `Evidence` of a byzantine validator in `BeginBlock` request. Before jailing any account because of
+byzantine fault, that evidence should be verified. Also, it should be checked that evidence provided by tendermint is
+not older than `MAX_EVIDENCE_AGE` in tendermint.
+:::
+
+### Jailing
+
+A validator is jailed if:
+
+1. They are not **live**, i.e., they failed to sign `MISSED_BLOCK_THRESHOLD` blocks in last
+   `BLOCK_SIGNING_WINDOW` blocks successfully. 
+1. They make a byzantine fault, e.g., they sign messages at same height and round.
+
+When a validator gets jailed, they cannot perform any operations relating to their account, for example,
+`withdraw_stake`, `deposit_stake`, `unbond_stake`, etc., until they are un-jailed. Also, a validator cannot be un-jailed
+before `account.jailed_until` which is set to `block_time + UNBONDING_PERIOD` while jailing. `UNBONDING_PERIOD` is a
+network parameter which can be configured during genesis.
+
+:::tip Important:
+`block_time` used in calculating `account.jailed_until` should be the time of the block at which the validator made the
+fault, which can be obtained from `evidence` provided by tendermint (current block for liveness faults).
+:::
+
+#### Un-jailing and Re-joining 
+
+When a jailed validator wishes to resume normal operations (after `account.jailed_until` has passed and the account is
+slashed), they can create `UnjailTx` which marks them as un-jailed. After successful un-jailing, validators can submit a
+`NodeJoinTx`, which will add them back to validator set.
+
+:::tip Important:
+`UnjailTx` should only be valid after funds are slashed from jailed account.
+:::
+
+### Slashing
+
+Validators are responsible for signing or proposing block at each consensus round. It is important that they maintain
+excellent availability and network connectivity to perform these tasks. A penalty performed by the slashing module
+should be imposed on validators' misbehavior or unavailability to reinforce this. Similar to jailing, a validator is
+slashed if:
+
+1. They are not **live**, i.e., they failed to sign `MISSED_BLOCK_THRESHOLD` blocks in last `BLOCK_SIGNING_WINDOW`
+   blocks successfully.
+1. They make a byzantine fault, e.g., they sign messages at same height and round.
+
+Unlike jailing, which happens immediately after punishments are triggered, slashing happens after `UNBONDING_PERIOD`.
+`UNBONDING_PERIOD` is a network parameter and can be configured during genesis. Validators are not immediately slashed
+because evidence for more faulty may be discovered after some time. If a validator makes multiple faults in
+`UNBONDING_PERIOD`, they'll only be slashed once for the worst fault in that period.
+
+:::tip Implementation note:
+It should be enforced in implementation that un-jailing can only be done after an account is slashed. So, even if an
+account can be un-jailed after `UNBONDING_PERIOD`, it should not be allowed to un-jail until it has been slashed.
+:::
+
+:::tip Important:
+A validator should not be slashed more than once within `UNBONDING_PERIOD`. If we get multiple slash requests for a
+validator within `UNBONDING_PERIOD`, we should ignore all requests other than the first one.
+:::
+
+#### Slashing Rate
+
+Whenever a validator is slashed, a percentage of their `bonded` and `unbonded` amount is transferred to `rewards_pool`.
+There are many factors involved in determining the slashing rate for a validator:
+
+1. `LIVENESS_SLASH_PERCENT` and `BYZANTINE_SLASH_PERCENT` are the two network parameters used while calculating the
+   slashing rate. Both of these can be configured in the genesis.
+1. `validator_voting_percent` is the voting percent of faulty validator in block `H` (`H` is the height of the block
+   when the validator made the fault).
+1. List of all the faulty validators in the block `H` and their `validator_voting_percent`s (according to above
+   definition).
+
+The algorithm for calculating `slashing_rate` is as follows:
+
+```
+validator_slash_percent = max(slash percent of individual faults)
+```
+
+For example, if a validator made a byzantine fault as well as they are non-live, then,
+
+```
+validator_slash_percent = max(liveness_slash_percent, byzantine_slash_percent)
+```
+
+And if there are `n` faulty validators in this period, then,
+
+```
+slashing_rate = 
+    validator_slash_percent * 
+    (
+        sqrt(validator_voting_percent_1) +
+        sqrt(validator_voting_percent_2) +
+        .. + 
+        sqrt(validator_voting_percent_n)
+    )^2
+```
+
+So, if one validator of 10% voting power faults, it gets a 10% slash (assuming `liveness_slash_percent` and
+`byzantine_slash_percent` are both 100%). While, if two validators of 5% voting power each fault together, they both get
+a 20% slash.
+
+```
+slashing_rate = max(liveness_slash_percent, byzantine_slash_percent) * (sqrt(validator_voting_percent_1) + sqrt(validator_voting_percent_2))^2
+              = max(1, 1) * (sqrt(0.05) + sqrt(0.05))^2                // assuming liveness_slash_percent and byzantine_slash_percent are both 100%
+              = 0.2
+```
--- a/docs/getting-started/punishments.md
+++ b/docs/getting-started/punishments.md
@@ -101,9 +101,9 @@ account can be un-jailed after `UNBONDING_PERIOD`, it should not be allowed to u
 :::
 
 :::tip Important:
-A validator should not be slashed more than once within `UNBONDING_PERIOD`. I a validator commits multiple faults before
-`account.jailed_until`, it should only be slashed with the highest slash amount in that period (can be calculated using
-below algorithm).
+A validator should not be slashed more than once within `UNBONDING_PERIOD`. If a validator commits multiple faults
+before `account.jailed_until`, it should only be slashed with the highest slash amount in that period (can be calculated
+using below algorithm).
 :::
 
 #### Slashing Rate
--- a/docs/getting-started/punishments.md
+++ b/docs/getting-started/punishments.md
@@ -41,8 +41,8 @@ be marked as **non-live** if they fail to successfully sign at least `50` blocks
 
 - Byzantine Faults (Double signing)
 
-    A validator is said to make a byzantine fault when they sign conflicting transactions at the same height and round.
-    Tendermint has mechanisms to publish evidence of validators that signed conflicting votes (it passes this 
+    A validator is said to make a byzantine fault when they sign conflicting messages/blocks at the same height and
+    round. Tendermint has mechanisms to publish evidence of validators that signed conflicting votes (it passes this 
     information to ABCI application in `BeginBlock` request), so they can be punished by the application.
 
 :::tip Implementation note:
--- a/docs/getting-started/punishments.md
+++ b/docs/getting-started/punishments.md
@@ -101,7 +101,7 @@ account can be un-jailed after `UNBONDING_PERIOD`, it should not be allowed to u
 :::
 
 :::tip Important:
-A validator should not be slashed more than once within `UNBONDING_PERIOD`. If a validator commits multiple faults
+A validator should not be slashed more than once within `UNBONDING_PERIOD` after they were jailed. If a validator commits multiple faults
 before `account.jailed_until`, it should only be slashed with the highest slash amount in that period (can be calculated
 using below algorithm).
 :::
--- a/docs/getting-started/punishments.md
+++ b/docs/getting-started/punishments.md
@@ -84,7 +84,7 @@ slashed), they can create `UnjailTx` which marks them as un-jailed. After succes
 Validators are responsible for signing or proposing block at each consensus round. It is important that they maintain
 excellent availability and network connectivity to perform these tasks. A penalty performed by the slashing module
 should be imposed on validators' misbehavior or unavailability to reinforce this. Similar to jailing, a validator is
-slashed if:
+slashed if any one of the following applies:
 
 1. They are not **live**, i.e., they failed to sign `MISSED_BLOCK_THRESHOLD` blocks in last `BLOCK_SIGNING_WINDOW`
    blocks successfully.
--- a/docs/getting-started/punishments.md
+++ b/docs/getting-started/punishments.md
@@ -101,9 +101,9 @@ account can be un-jailed after `UNBONDING_PERIOD`, it should not be allowed to u
 :::
 
 :::tip Important:
-A validator should not be slashed more than once within `UNBONDING_PERIOD` after they were jailed. If a validator commits multiple faults
-before `account.jailed_until`, it should only be slashed with the highest slash amount in that period (can be calculated
-using below algorithm).
+A validator should not be slashed more than once within `UNBONDING_PERIOD` after they were jailed. If a validator
+commits multiple faults before `account.jailed_until`, it should only be slashed with the highest slash amount in that
+period (can be calculated using below algorithm).
 :::
 
 #### Slashing Rate
--- a/docs/getting-started/punishments.md
+++ b/docs/getting-started/punishments.md
@@ -55,8 +55,13 @@ not older than `MAX_EVIDENCE_AGE` in tendermint.
 ### Inactivity Punishment
 
 When a validator fails to successfully sign `MISSED_BLOCK_THRESHOLD` blocks in last `BLOCK_SIGNING_WINDOW` blocks, it is
-immediately punished by deducting funds from their bonded and unbonded amount. The funds to be deducted are calculated
+immediately punished by deducting funds from their bonded and unbonded amount and removing them from active validator
+set if the resulting bonded amount is still greater than minimum required stake. The funds to be deducted are calculated
 based on `LIVENESS_SLASH_PERCENT`.
+
+:::tip Note:
+The validator is not **jailed** in this scenario. They can immediately send a `NodeJoinTx` to join back as a validator.
+:::
 
 ### Jailing
 
--- a/docs/getting-started/punishments.md
+++ b/docs/getting-started/punishments.md
@@ -16,8 +16,9 @@ parameters and their effect on behavior of validator punishments is discussed la
 1. `BLOCK_SIGNING_WINDOW`: Number of blocks for which the moving average is calculated for uptime tracking.
 1. `MISSED_BLOCK_THRESHOLD`: Maximum number of blocks with faulty/missed validations allowed for an account in last
    `BLOCK_SIGNING_WINDOW` blocks before it gets jailed.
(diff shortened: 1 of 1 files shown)


docs/getting-started/punishments.md
```diff
@@ -0,0 +1,148 @@
...
+
+#### Slashing Rate
+
+Whenever a validator is slashed, a percentage of their `bonded` and `unbonded` amount is transferred to `rewards_pool`.
+There are many factors involved in determining the slashing rate for a validator:
+
+1. `LIVENESS_SLASH_PERCENT` and `BYZANTINE_SLASH_PERCENT` are the two network parameters used while calculating the
+   slashing rate. Both of these can be configured in the genesis.
+1. `validator_voting_percent` is the voting percent of faulty validator in block `H - 1` (`H` is the height of the block
+   when the validator made the fault).
```
yihuang: I think if `H` is the height of block when the validator made the fault, then the block height when the fault be detected is within range `(H, H + max_evidence_age)`,

devashishdxt: That is correct. We can get evidence of a fault later, but the `block_height` in evidence will be the actual block height at which the validator made fault (`H`). `validator_voting_percent` at `H` should be used for `slash_ratio` calculations.

Edit: Changed `H - 1` to `H`.

devashishdxt: Edit: Changed `H - 1` to `H`.

yihuang: - So we need to store history voting powers for each block.
- The `total_voting_power` is still need to be computed based on the current state? Because multiple evidences meight have different block heights?

devashishdxt: No. `total_voting_power` is also of block `H`. I there are multiple evidences for same validator, they'll only be slashed once if difference in their block times is less than `UNBONDING_PERIOD`.

devashishdxt: >     * So we need to store history voting powers for each block.

Yes. We'll have to store history until `UNBONDING_PERIOD`.

yihuang:
> How it prevents bugs like crypto-com/chain#1079?
> 
> This PR adds a necessary condition that the account must be already slashed before it can be unjailed.

I think it still needs to find the source of that bug because it's not supposed to happen if `jail_duration > slash_wait_period`.

devashishdxt:
> > How it prevents bugs like [crypto-com/chain#1079](https://github.com/crypto-com/chain/issues/1079)?
> > This PR adds a necessary condition that the account must be already slashed before it can be unjailed.
> 
> I think it still needs to find the source of that bug because it's not supposed to happen if `jail_duration > slash_wait_period`.

That's why I used the word **prevents**. 😄 
Also, this PR removes `JAIL_DURATION` and `SLASH_WAIT_PERIOD` network parameters.

docs/getting-started/punishments.md
```diff
@@ -0,0 +1,148 @@
...
+parameters and their effect on behavior of validator punishments is discussed later in this document.
+
+1. `UNBONDING_PERIOD`: Unbonding period will be used as jailing period (time for which an account is jailed after it
+   gets punished) and also as slashing period (time to wait before slashing funds from an account). This should be
+   greater than or equal to `MAX_EVIDENCE_AGE` in tendermint.
+1. `BLOCK_SIGNING_WINDOW`: Number of blocks for which the moving average is calculated for uptime tracking.
+1. `MISSED_BLOCK_THRESHOLD`: Maximum number of blocks with faulty/missed validations allowed for an account in last
+   `BLOCK_SIGNING_WINDOW` blocks before it gets jailed.
+1. `LIVENESS_SLASH_PERCENT`: Percentage of funds (bonded + unbonded) slashed when validator is not live.
+1. `BYZANTINE_SLASH_PERCENT`: Percentage of funds (bonded + unbonded) slashed when validator makes a byzantine fault.
```
tomtau: I wonder whether there is any rigorous justification for having two rates.
Having only one would make the design a bit simpler (one less parameter, slash can happen immediately).
The only justification I've seen was along the lines "punishing non-live validator should be less harsh", but I wonder whether there's some more detailed rigorous argument behind that (e.g. based on data gathered from some existing network operations)

docs/getting-started/punishments.md
```diff
@@ -0,0 +1,148 @@
...
+    supported). Tendermint passes signing information to ABCI application as `last_commit_info` in `BeginBlock` request.
+
+:::tip Example:
+For example, if `BLOCK_SIGNING_WINDOW` is `100` blocks and `MISSED_BLOCK_THRESHOLD` is `50` blocks, a validator will
+be marked as **non-live** if they fail to successfully sign at least `50` blocks in last `100` blocks.
+:::
+
+- Byzantine Faults (Double signing)
+
+    A validator is said to make a byzantine fault when they sign conflicting transactions at the same height and round.
```
tomtau: > conflicting transactions

perhaps conflicting messages/blocks? there could be two empty blocks for the same height that cause different outcomes (staked states rewarded, punished,...)

docs/getting-started/punishments.md
```diff
@@ -0,0 +1,148 @@
...
+period of time, collect rewards etc.
+
+## Network Parameters
+
+Below are all the network parameters used to configure the behavior of validator punishments. Details of all these
+parameters and their effect on behavior of validator punishments is discussed later in this document.
+
+1. `UNBONDING_PERIOD`: Unbonding period will be used as jailing period (time for which an account is jailed after it
+   gets punished) and also as slashing period (time to wait before slashing funds from an account). This should be
+   greater than or equal to `MAX_EVIDENCE_AGE` in tendermint.
```
lezzokafka, line 15: In the design of tendermint, setting a long `UNBONDING_PERIOD` is a way to prevent the [long range attack](https://github.com/cosmos/cosmos/blob/master/WHITEPAPER.md#preventing-long-range-attacks). 
In this case, it could be a very harsh punishment for the validator who commits a liveness fault. 

For example, If the `UNBONDING_PERIOD` (so does the jailing period in this setting) is 21 days (as in Cosmos), the total punishment for the validator will be:
- the liveness slashing (by percentage) and 
- the opportunity cost of missing the validator reward during the jailing period, which will be 3/52 (5.72%) of the yearly reward.

docs/getting-started/punishments.md
tomtau: also one note that there may be one other attributable byzantine fault: https://github.com/cosmos/cosmos-sdk/blob/master/x/slashing/spec/07_tombstone.md#single-slashing-amount
>It is currently planned to include the following fault in the near future:
> Signing a precommit when you're in unbonding phase (needed to make light client bisection safe)

docs/getting-started/punishments.md
devashishdxt: How does cosmos punishes non live validators? 

Also, if `UNBONDING_PERIOD` is high, `BLOCK_SIGNING_WINDOW` and `MISSED_BLOCK_THRESHOLD` should also be high enough (for example, 10 days of `BLOCK_SIGNING_WINDOW` and 5 days of `MISSED_BLOCK_THRESHOLD` converted into number of blocks) so that validator does not get jailed for small liveness issues.

docs/getting-started/punishments.md
tomtau: or there's a proposal of having that in ABCI: https://github.com/tendermint/tendermint/blob/master/docs/architecture/adr-045-abci-evidence.md

docs/getting-started/punishments.md
lezzokafka: Yes, setting a high `BLOCK_SIGNING_WINDOW` and `MISSED_BLOCK_THRESHOLD` will be a solution I would give as well. 

In cosmos network (hub-2), a validator will be jailed for 10 mins and slashed by 0.01% of the total stake  If it misses 10000*0.05 = 500 blocks in the last 10,000 blocks. That means (10000 - 500)*6 / 3600 ~ 15.8 Hrs of total downtime given blocks being generated every 6 seconds.  

As a reference, here’s the average liveness failure rate for the first 2.48 million blocks in the cosmos network (hub-2).

Failure Type | Average failure rate 
-- | --
Precommits missed as a block proposer | 0.29%
Precommits missed as a validator | 0.26%

docs/getting-started/punishments.md
```diff
@@ -0,0 +1,148 @@
...
+#### Slashing Rate
+
+Whenever a validator is slashed, a percentage of their `bonded` and `unbonded` amount is transferred to `rewards_pool`.
+There are many factors involved in determining the slashing rate for a validator:
+
+1. `LIVENESS_SLASH_PERCENT` and `BYZANTINE_SLASH_PERCENT` are the two network parameters used while calculating the
+   slashing rate. Both of these can be configured in the genesis.
+1. `validator_voting_percent` is the voting percent of faulty validator in block `H` (`H` is the height of the block
+   when the validator made the fault).
+1. List of all the faulty validators in the block `H` and their `validator_voting_percent`s (according to above
```
yihuang: Isn't `H` is different for each evidence detected in the current block, maybe you mean current block here?

devashishdxt: No. `H` is the height of the block where the fault was committed. Not the current block height.

docs/getting-started/punishments.md
```diff
@@ -0,0 +1,148 @@
...
+:::tip Important:
+`block_time` used in calculating `account.jailed_until` should be the time of the block at which the validator made the
+fault, which can be obtained from `evidence` provided by tendermint (current block for liveness faults).
+:::
+
+#### Un-jailing and Re-joining 
+
+When a jailed validator wishes to resume normal operations (after `account.jailed_until` has passed and the account is
+slashed), they can create `UnjailTx` which marks them as un-jailed. After successful un-jailing, validators can submit a
+`NodeJoinTx`, which will add them back to validator set.
```
tomtau: Given the re-working of https://docs.tendermint.com/master/spec/consensus/light-client/
https://github.com/tendermint/tendermint/blob/master/docs/architecture/adr-044-lite-client-with-weak-subjectivity.md
https://github.com/tendermint/tendermint/blob/master/docs/architecture/adr-046-light-client-implementation.md

I'm not sure how safe the light client is if it doesn't go online for more than the unbonding period -- @yihuang can comment or we can ask on TM Slack.
Will banning the consensus pub keys that were attributed a Byzantine failure help (so the node re-join needs to be with a different one)? So that the light client can assume that it won't / shouldn't see repeated blocks proposed by pubkey that's been attributed a byzantine fault ?

docs/getting-started/punishments.md
devashishdxt: So, in Cosmos, a validator will be slashed multiple times if it commits both, liveness fault as well as byzantine fault within `MAX_EVIDENCE_AGE`?

The reason we consider liveness faults with byzantine faults is because we don't want to punish validator multiple times for different faults.

lezzokafka: >`MISSED_BLOCK_THRESHOLD` should also be high enough (for example, 10 days of `BLOCK_SIGNING_WINDOW` and 5 days of `MISSED_BLOCK_THRESHOLD` converted into number of blocks) so that validator does not get jailed for small liveness issues.


Actually, the conversion could be something tricky; to do this, we would need to play with the tendermint setting to get some expected average block time?

devashishdxt: > > `MISSED_BLOCK_THRESHOLD` should also be high enough (for example, 10 days of `BLOCK_SIGNING_WINDOW` and 5 days of `MISSED_BLOCK_THRESHOLD` converted into number of blocks) so that validator does not get jailed for small liveness issues.
> 
> Actually, the conversion could be something trickly; to do this, we would need to play with the tendermint setting to get some expected average block time?

Or we can change the unit of measurment of `BLOCK_SIGNING_WINDOW` and `MISSED_BLOCK_THRESHOLD` to `Duration (in secs)` from `number of blocks`. But, this will make it very tricky to calculate liveness.

docs/getting-started/punishments.md
tomtau: By history, do you mean just one entry per faulting validator (the voting power / total voting power at the time of detection?)?

tomtau (commented):
left some comments

docs/getting-started/punishments.md
devashishdxt: We'll have to store following data for each block until `block_time + UNBONDING_PERIOD`:

```
{
    "block_time",
    "block_height",
    "block_validators": [{
        "validator_address",
        "validator_voting_power",
    }],
    "faulty_validators_in_block": ["validator_address_1", "validator_address_2", ..],
}
```

When we get an evidence of a fault committed at `block_height`, we'll add faulty validator in above data. We can delete this data when funds are slashed for faulty validators in this block (which should be `block_time + UNBONDING_PERIOD`).

devashishdxt: Added a note:
```
A validator should not be slashed more than once within `UNBONDING_PERIOD`. If a validator commits multiple faults before
`account.jailed_until`, it should only be slashed with the highest slash amount in that period (can be calculated using
below algorithm).
```

docs/getting-started/punishments.md
devashishdxt: I agree. Having only one `SLASH_PERCENT` will be much more simpler. @lezzokafka Any comments?

docs/getting-started/punishments.md
devashishdxt: Changed.

docs/getting-started/punishments.md
yihuang: > No. `H` is the height of the block where the fault was committed. Not the current block height.

When there are multiple evidences, each one might have a different block height, what does this mean: "all the faulty validators in the block `H`".

devashishdxt: Each `BeginBlock` contains some evidences of faults made by validators. Each evidence has a `block_height` associated with it which is the height of block at which the fault was committed. If a validator commits multiple faults at different block heights (which gets detected later), its `slash_amount` will be calculated for each block (where it made a fault) and only the maximum of them will be slashed.

devashishdxt: > all the faulty validators in the block `H`

This means all the validators who committed fault in block `H`. It may get detected later in different block heights.

docs/getting-started/punishments.md
yihuang: In tendermint spec, light client has a parameter `TRUSTED_PERIOD < UNBONDING_PERIOD`, it's not implemented in our master, but it's already in `tendermint-rs`'s master.

docs/getting-started/punishments.md
lezzokafka: Besides the common justification, the consideration to me is more about the intention. Basically the logic is that we should be forgetful and show the mercy to those unintentional mistakes (misconfigurations etc.) but be harsh on the intentional one. This makes a different level of punishment.

The [only](https://twitter.com/zmanian/status/1145072296723275776?lang=en) double sign fault on was due to misconfigurations (at least this is what they claimed). By given key management tools like kms, it should be relatively hard for someone to commit such a fault by accident with a proper setup. Therefore, if you happened to break the rules, you must be up to something, and I would have to consider you to be malicious and put a severe punishment on you. 

Anyways, as I [mentioned](https://github.com/crypto-com/chain-docs/pull/89#discussion_r381098948) earlier, if the jailing period is very long and we put a high tolerance level for the liveness fault as suggested by @devashishdxt, it would end up doing the same task by having just one `SLASH_PERCENT`.

docs/getting-started/punishments.md
lezzokafka: > Or we can change the unit of measurment of `BLOCK_SIGNING_WINDOW` and `MISSED_BLOCK_THRESHOLD` to `Duration (in secs)` from `number of blocks`. But, this will make it very tricky to calculate liveness.

Guess It would make it even more complicated! 👯‍♂ 

But let's take a step back and assume that we find an excellent way to do this conversion, having a high threshold would also give you another drawback to the network: There could be validators hanging around, doing nothing for days without any consequence..!

devashishdxt: > But let's take a step back and assume that we find an excellent way to do this conversion, having a high threshold would also give you another drawback to the network: There could be validators hanging around, doing nothing for days without any consequence..!

Is there any problem with this? If they're not doing anything, they won't get rewarded.

docs/getting-started/punishments.md
devashishdxt: Removed two different network parameters. Also changed parameters name to `SLASH_RATE_MULTIPLIER` which is more appropriate.

docs/getting-started/punishments.md
lezzokafka: > Is there any problem with this? If they're not doing anything, they won't get rewarded.

Sure, by consequence, I was referring to slashing. In this case, there could be some _inactive_ staking, which has been staked, but not really participating in the consensus for days. 
Of course, they would not receive the reward if they propose no block, and it should be the validator responsibility to make sure it runs well if they want to be rewarded.

docs/getting-started/punishments.md
```diff
@@ -0,0 +1,148 @@
...
+:::tip Important:
+`UnjailTx` should only be valid after funds are slashed from jailed account.
+:::
+
+### Slashing
+
+Validators are responsible for signing or proposing block at each consensus round. It is important that they maintain
+excellent availability and network connectivity to perform these tasks. A penalty performed by the slashing module
+should be imposed on validators' misbehavior or unavailability to reinforce this. Similar to jailing, a validator is
+slashed if:
```
lezzokafka:
```suggestion
slashed if any one of the following applies:
```

docs/getting-started/punishments.md
```diff
@@ -0,0 +1,148 @@
...
+
+:::tip Implementation note:
+Tendermint passes `Evidence` of a byzantine validator in `BeginBlock` request. Before jailing any account because of
+byzantine fault, that evidence should be verified. Also, it should be checked that evidence provided by tendermint is
+not older than `MAX_EVIDENCE_AGE` in tendermint.
+:::
+
+### Jailing
+
+A validator is jailed if:
```
lezzokafka:
```suggestion
A validator is jailed if any one of the following applies:
```

docs/getting-started/punishments.md
```diff
@@ -0,0 +1,148 @@
...
+because evidence for more faulty may be discovered after some time. If a validator makes multiple faults in
+`UNBONDING_PERIOD`, they'll only be slashed once for the worst fault in that period.
+
+:::tip Implementation note:
+It should be enforced in implementation that un-jailing can only be done after an account is slashed. So, even if an
+account can be un-jailed after `UNBONDING_PERIOD`, it should not be allowed to un-jail until it has been slashed.
+:::
+
+:::tip Important:
+A validator should not be slashed more than once within `UNBONDING_PERIOD`. If a validator commits multiple faults
```
lezzokafka, line 99:
```suggestion
A validator should not be slashed more than once within `UNBONDING_PERIOD` after they were jailed. If a validator commits multiple faults
```

docs/getting-started/punishments.md
yihuang: > > all the faulty validators in the block `H`
> 
> This means all the validators who committed fault in block `H`. It may get detected later in different block heights.

So is it only calculated at `H + max_evidence_age`, that's when "all the validators who committed fault in block `H`" get revealed? 
I'm still confused, maybe write down the computation process in the docs?

lezzokafka (commented):
Some more besides the [discussion](https://github.com/crypto-com/chain-docs/pull/89/files#r381098948) earlier

docs/getting-started/punishments.md
lezzokafka: Further reading:

There were proposal suggesting to rise the double sign slashing ratio from 5% to 20%, [here](https://forum.cosmos.network/t/proposal-draft-suggested-changes-to-a-few-cosmological-constants/1984)  is some of the discussion.

docs/getting-started/punishments.md
devashishdxt: @tomtau @yihuang How will it affect our punishement specifications?

docs/getting-started/punishments.md
devashishdxt: @yihuang Do you have any more questions/concerns around this? If no, we can resolve this conversation.

docs/getting-started/punishments.md
tomtau: my understanding is that permanently revoking consensus pub keys that've been attributed a byzantine fault may be required, otherwise (infrequently synchronised) light clients may have hard times due to long range attacks and the like.

https://pkg.go.dev/github.com/tendermint/tendermint/lite2?tab=doc

devashishdxt: Ok. Added a note to add public keys of byzantine validators in a list of permanently banned public keys.

docs/getting-started/punishments.md
tomtau: > Is there any problem with this? If they're not doing anything, they won't get rewarded.

 yes, lower performance due to timeouts on many offline validators.

docs/getting-started/punishments.md
yihuang: > @yihuang Do you have any more questions/concerns around this? If no, we can resolve this conversation.

I still don't fully understand the logic here, especially the computation steps and timings.

docs/getting-started/punishments.md
tomtau: Note I'm not sure whether that's required or enough (I wrote "may be required")! that's why I recommended studying more the current light client specs/implementation assumptions or asking on TM Slack

docs/getting-started/punishments.md
devashishdxt: @lezzokafka @tomtau Do you have any suggestions on the behaviour when a validator is non-live?

docs/getting-started/punishments.md
tomtau: If it's required, like all other stuff, it should have some justification -- docs should point to a resource

docs/getting-started/punishments.md
devashishdxt: What part are you finding difficult to understand? Maybe there is a scope to simplify the language.

docs/getting-started/punishments.md
tomtau: Since there's only one rate -- https://github.com/crypto-com/chain-docs/pull/89/files#diff-8f9dcaf1e4318fb72804d9070866d362R19

can't this be simplified? just slash directly at that block? No need for storing anything

devashishdxt: This is required because it may take some time before we find all the evidences of faults committed by different validators in a block. And, because we're slashing validators based on faults made by other validators in the same block (proportional slashing), we need to store this information for an extended period. Having only single rate will not simplify this. This can be simplified if we give up the idea of proportional slashing and slash validators irrespective of faults made by other valdiators in that block.

docs/getting-started/punishments.md
yihuang: > What part are you finding difficult to understand? Maybe there is a scope to simplify the language.

For example, validator `A` performed a byzantine fault at block `H`, which is detected at `H+n, n < max_evidence_age`. So what happens at block `H+n`, and what happens at block `H+max_evidence_age`?

docs/getting-started/punishments.md
tomtau: I recommend studying existing mechanism designs and having a setup that wouldn't lead to performance degrades due to misaligned incentives

docs/getting-started/punishments.md
tomtau: If it's defined per-block (that's my interpretation "in_block"), it's detected immediately with liveness and impossible with double-signs (only one leader per block -- only one fault attribution/source for that block)?

docs/getting-started/punishments.md
devashishdxt: Here is the whole process:
- Block `H`: Validator `A` commits a byzantine fault.
- Block `H + n`: We get he evidence of that fault in `BeginBlock` request. (`n < max_evidence_age`). We do following things at this point:
  - We jail this validator and set `jailed_until = current_block_time + UNBONDING_PERIOD`. Here `current_block_time` is the `block_time` of block `H + n`.
  - We add this validator in the list of validators who committed a fault at block `H`. This list also contains all the other validators who made fault in the same block.
  - We calculate the new `slash_amount` and update all the validators who made fault in block `H`.
- Block `H + n + UNBONDING_PERIOD`: We slash the validator based on the `slash_amount` stored and the validator can now be unjailed in future blocks.

docs/getting-started/punishments.md
devashishdxt: > If it's defined per-block (that's my interpretation "in_block"), it's detected immediately with liveness and impossible with double-signs (only one leader per block -- only one fault attribution/source for that block)?

I'm a bit confused. Are you saying that only one validator can make byzantine fault per block?

tomtau: Ignore my comment -- it's attributed from signatures on `Precommit`, not `Propose`

docs/getting-started/punishments.md
yihuang: Yeah, that's what I need, and I think it should be added to the docs.

> We calculate the new `slash_amount` and update all the validators who made fault in block `H`.

So the `slash_amount` is not finalized until `H + max_evidence_age` right, as when there is new fault committed at block `H` get detected, it will be updated.

devashishdxt: That is correct.

docs/getting-started/punishments.md
```diff
@@ -0,0 +1,153 @@
...
+Below are all the network parameters used to configure the behavior of validator punishments. Details of all these
+parameters and their effect on behavior of validator punishments is discussed later in this document.
+
+1. `UNBONDING_PERIOD`: Unbonding period will be used as jailing period (time for which an account is jailed after it
+   gets punished) and also as slashing period (time to wait before slashing funds from an account). This should be
+   greater than or equal to `MAX_EVIDENCE_AGE` in tendermint.
+1. `BLOCK_SIGNING_WINDOW`: Number of blocks for which the moving average is calculated for uptime tracking.
+1. `MISSED_BLOCK_THRESHOLD`: Maximum number of blocks with faulty/missed validations allowed for an account in last
+   `BLOCK_SIGNING_WINDOW` blocks before it gets jailed.
+1. `SLASH_RATE_MULTIPLIER`: Multiplier of funds (bonded + unbonded) slashed when validator makes a fault (liveness fault
```
yihuang: I think we should provide a set of typical config values for production, so we can do some basic calculations based on it, and understand the logic more intuitively.

devashishdxt: I'm on it! I'm writing a small simulation program to simulate different scenarios and find the outcome.

tomtau (commented):
is this finalised or what parts are missing?

devashishdxt:
> is this finalised or what parts are missing?

There is still some scope for improvement when validators make liveness faults and gets punished. As @lezzokafka mentioned in [this](https://github.com/crypto-com/chain-docs/pull/89#discussion_r381098948) comment, the current way of punishing validators for liveness faults in very harsh given that livenss faults are not very rare and most of the times, they are unintentional (https://github.com/crypto-com/chain-docs/pull/89#discussion_r381104675).

devashishdxt:
> > is this finalised or what parts are missing?
> 
> There is still some scope for improvement when validators make liveness faults and gets punished. As @lezzokafka mentioned in [this](https://github.com/crypto-com/chain-docs/pull/89#discussion_r381098948) comment, the current way of punishing validators for liveness faults in very harsh given that livenss faults are not very rare and most of the times, they are unintentional ([#89 (comment)](https://github.com/crypto-com/chain-docs/pull/89#discussion_r381104675)).

@tomtau @lezzokafka @yihuang Made some changes for punishing inactive validators differently. Byzantine validators are still slashed proprtionally (based on other validators that commit byzantine fault in the same block).

docs/getting-started/punishments.md
```diff
@@ -0,0 +1,157 @@
...
+There are many factors involved in determining the slashing rate for a validator:
+
+1. `BYZANTINE_SLASH_RATE_MULTIPLIER` is the network parameters used while calculating the slashing rate. It can be
+   configured in the genesis.
+1. `validator_voting_percent` is the voting percent of faulty validator in block `H` (`H` is the height of the block
+   when the validator made the fault).
+1. List of all the faulty validators in the block `H` and their `validator_voting_percent`s (according to above
+   definition).
+
+The algorithm for calculating `slashing_rate` when there are `n` validators who committed faults in a block:
```
yihuang: With newer modifications, I think proportion should be calculated based on fault detect time, rather than commit time?

yihuang (approved):
The final version is much simpler, I guess simple is better,
and we can always improve further in the future when we know better.

lezzokafka (approved):
Looking good! 
Thanks for the great effort for this PR,
will perform further analysis and fine-tune the parameters based on this. 👍

lezzokafka:
Btw, you might want to use `::: warning` for those important notes?

Issue #81, opened 2020-02-13
Problem: reward<->punishment<->monetary update interactions not explicitly stated


tomtau:
there should be a dedicated specification page

tomtau:
"monetary update interactions" == transactions manipulating bonded amount (deposit_tx and unbond_tx)

tomtau:
some early WIP: https://github.com/crypto-com/chain-docs/pull/75

yihuang:
Also:
- How to choose active validators
- What's the status of validator who joined but not chosen into the top `max_validators`
- Liveness tracking of them
- Store validator metadata in an in-memory data structure or in the `StakedState` which is stored in a merkle trie. (I prefer the former, only let the merkle trie to handle the basic account model)

yihuang:
We also need to clarify the nonce somewhere, which stats change should increase the nonce:
- bonded/unbonded change, definitely yes.
- council node metadata change?
- jail/unjail ?
- liveness tracker change, I guess no?

tomtau:
> Also:
> 
> * How to choose active validators
> * What's the status of validator who joined but not chosen into the top `max_validators`
> * Liveness tracking of them
> * Store validator metadata in an in-memory data structure or in the `StakedState` which is stored in a merkle trie. (I prefer the former, only let the merkle trie to handle the basic account model)

yep, all of these are a part of this.
from the previous discussions / sketches:

 > How to choose active validators

non-punished, ordered by bonded amount, capped by max network parameter

> What's the status of validator who joined but not chosen into the top `max_validators`

reaching "max" isn't expected, so as always simplicity is preferred -- they'll just be removed without any punishments and need a manual action to rejoin (e.g. deposit more coins)... some of their data may need to be kept around for the unbonding period duration after they are gone

> Liveness tracking of them

more details needed in https://github.com/crypto-com/chain-docs/blob/master/docs/getting-started/punishments.md ?

> Store validator metadata in an in-memory data structure ...

it depends on what metadata -- one other consideration here is that this is potentially untrusted input, so if this metadata can grow by actions done by operators, it's a potential vector for DoS.
e.g. the pubkeys of byzantine actors don't need to be in merkle trie, but should be at least persisted in local kv-storage.

> We also need to clarify the nonce somewhere, which stats change should increase the nonce:
> 
> * council node metadata change?

yes
> * jail/unjail ?

yes
> * liveness tracker change, I guess no?

no

yihuang:
[...]

PR #94, opened 2020-02-28
Problem (Fix #82): reward mechanisms not defined precisely

### LINKED ISSUE #82: Problem: reward mechanisms not defined precisely
need fairness explanation / update how votes are processed
--
define precisely (somewhat "formally") in relation to Tendermint ABCI what each part does, so that this can be easily checked against the implementation:
what are data inputs from where/when
how are they processed
what are outputs and when they are emitted

yihuang:
Solution:
- Add reward mechanisms in terms of abci events and rust code

### DIFF
--- a/docs/modules/rewarding.md
+++ b/docs/modules/rewarding.md
@@ -1,0 +1,389 @@
+# Rewarding
+
+## Abstract
+
+Rewarding module enable crypto chain to reward protocol participators, the design should meet following requirements:
+
+- The total supply of CRO is fixed.
+- The distribution should be "fair", i.e. all honest participators should get rewards, non-participators should not, and the amount of rewards get distributed to a validator should be proportional to it's contribution (sum of the voting power of each vote).
+- The newly emited coins for rewarding is adjusted according to [total staking](#total-staking) and current block time.
+
+The basic procedure is:
+
+- Record participators for each block, sum the voting powers for each participator at the vote time. 
+- Distribute the accumulated reward pool according to the recorded sum powers at the end of reward period.
+
+For example, in one reward period with N blocks, validator `A` only participate in the first block, and all the other validators particiate in all blocks, assuming the validator set is static during this period, then we have:
+
+```
+p: voting power of validator A
+p': sum of voting power of all other validators
+
+block1: participators: p', p
+block2: participators: p'
+block3: participators: p'
+...
+blockN: participators: p'
+
+# The rewards of A is:
+reward_a = reward_pool * p / (p + p' * N)
+```
+
+The result of reward distribution should be written into events of block results.
+
+### Concepts
+
+#### Reward participator
+
+All the validators who follow the protocol honestly should participate in the reward distribution.
+
+In tendermint's terms, following the protocol honestly means being a signed voter in `last_commit_info` of `BeginBlockRequest`.
+
+It can be described precisely as:
+
+```rust
+begin_block_req
+  .last_commit_info
+  .votes
+  .filter(|vote| vote.signed_last_block)
+```
+
+#### Reward period
+
+Rewards are distributed to the validators periodically, the length of the period is configured in network parameter `reward_period_seconds` in genesis config.
+
+#### Reward pool
+
+During the reward period, the following sources of funds are accumulated into reward pool:
+
+- [Monetary expansion](#monetary-expansion)
+- [Transaction Fees](../getting-started/transaction.md#transaction-fee)
+- Slashings
+
+The reward pool get accumulated through the whole reward period, it get distributed once at the end of reward period.
+
+Because of integer calculations, there will be a few remained funds after distribution, they should be left to the next period.
+
+If there is zero participator for the period, the rewards pool should be left to the next period, rather than get burned.
+
+#### Total staking
+
+The sum of bonded coins of all active validators at some block.
+
+#### Monetary expansion
+
+New coins are minted into the reward pool at the end of reward period.
+
+The amount is calculated based on the total staking of current state and current block time.
+
+The total minted coins for rewarding shouldn't exceed the amount configured in network parameter `monetary_expansion_cap`.
+
+The calculation is as follows:
+
+```python
+R0 = network_parameter["monetary_expansion_r0"]  # upper bound for the reward rate p.a.
+P = network_parameter["reward_period_seconds"]
+tau = current tau  # refer to section "Decay of parameter tau"
+S = total staking  # refer to section "Total staking"
+Y = 365 * 24 * 60 * 60  # seconds of a year
+
+CAP = network_parameter["monetary_expansion_cap"]
+minted = total minted coins so far  # recorded in state
+
+R = (R0 / 1000) * exp(-S / tau)
+N = floor(S * (pow(1 + R, P / Y) - 1))
+N = N - N % 10000  # less precision requirement
+result = min(N, CAP - minted)  # the amount of newly minted coins
+```
+
+#### Decay of parameter tau
+
+The parameter `tau` in the monetary expansion calculation should be initialized to network parameter `monetary_expansion_tau` and gets decayed for each reward distribution. 
+
+The rate of decaying is configured in network parameter `monetary_expansion_decay`.
+
+The calculation is like this:
+
+```python
+rate = network_parameter["monetary_expansion_decay"]
+
+# init
+tau = network_parameter["monetary_expansion_tau"]
+
+# decay
+tau = tau * rate / 1000
+```
+
+To prevent overflow of integer multiplication, it can be transformed into:
+
+```python
+tau = (tau / 1000) * rate + (tau % 1000) * rate
+```
+
+#### Fixed point arithemetic
+
+We should use continued fraction method to compute the `exp` and `pow` with fixed point arithemetics.
+
+TODO
+
+## Implementation
+
+### Network parameters
+
+```rust
+pub struct RewardsParameters {
+    /// Reward period in seconds
+    pub reward_period_seconds: u64,
+
+    /// Maximum minted coins for rewards
+    pub monetary_expansion_cap: Coin,
+    /// Monetary expansion formula parameter R0
+    pub monetary_expansion_r0: Milli,
+    /// Monetary expansion formula parameter tau
+    pub monetary_expansion_tau: u64,
+
+    /// Decay rate of the parameter tau
+    pub monetary_expansion_decay: u64,
+}
+```
+
+Validation of `RewardsParameters`:
+
+```rust
+impl RewardsParameters {
+    pub fn validate(&self) -> Result<(), &'static str> {
+        if self.monetary_expansion_r0 > Milli::integral(1) {
+            return Err("R0 can't > 1");
+        }
+        if self.monetary_expansion_tau == 0 {
+            return Err("tau can't == 0");
+        }
+        if self.reward_period_seconds > 365 * 86400 {
+            return Err("reward period can't exceed 365 days");
+        }
+        if self.monetary_expansion_decay > 1_000_000 {
+            return Err("decay can't > 1_000_000");
+        }
+        Ok(())
+    }
+}
+```
+
+### State
+
+```rust
+/// Participate in the computation of app_hash, empty block should not modify
+pub struct RewardState {
+    /// Fees and slashings accumulated through current reward period
+    pub period_bonus: Coin,
+    /// last block height that updated it
+    pub last_block_height: BlockHeight,
+    /// last reward distribution time, default to the genesis time
+    pub last_distribution_time: Timespec,
+    /// Record how many coins have been minted so far, shouldn't exceed the monetary_expansion_cap
+    pub minted: Coin,
+    /// Current tau value of monetary expansion calculation
+    pub current_tau: u64,
+}
+
+pub struct ValidatorState {
+    ...
+
+    /// Records the participators for reward distribution, the value is the sum of voting powers for each signed vote of the participator
+    reward_participators: HashMap<StakingAddress, u64>,
+}
+
+pub struct ChainNodeApp {
+    pub reward_state_updated: bool
+}
+```
+
+Use `u64` for the sum of voting powers is enough, because total staking powers is smaller than 10 billion, so if maximum staking powers participates in all the blocks, the number of blocks we can support before `u64` overflows is:
+
+```
+>>> 9223372036854775807 / 10000000000
+922337203
+```
+
+It should be more than enough for one reward period, but still the implementation should avoid panic or wrap-around when overflows happen in extreme case.
+
+### Predefined symbols
+
+In the following sections, we use following predefined symbols:
+
+- `reward_state` for the current `RewardState`
+- `validator_state` for the current `ValidatorState`
+- `node_state` for the current `ChainNodeApp`
+- `network_parameter` for the current `NetworkParameter`
+
+### ABCI Events
+
+#### Begin block
+
+The main actions all happens in begin block event, it's the time we get reports of participators through `last_commit_info`, also the time the node state is consistent.
+
+1. Collect the validator addresses of reward participators from `last_commit_info`.
+2. Record them, and accumulate their voting powers.
+3. If a reward period has passed, distribute the reward pool according to the participator statistics.
+
+##### `begin_block`
+
+```rust
+fn begin_block(req: RequestBeginBlock) {
+    let addrs = req.last_commit_info.votes
+        .filter(|vote| vote.signed_last_block)
+        .map(|vote| get_validator_address(vote));
+
+    for addr in addrs {
+        // Panic or not is decided by the implementator
+        let staking_addr = find_staking_address(addr).unwrap();
+        // Panic or not is decided by the implementator
+        let power = get_voting_power(staking_addr).unwrap();
+        let amount = u64::from(power);
+        if validator_state.reward_participators.contains(staking_addr) {
+            // Use saturating_add for better robustness
+            validator_state.reward_participators[staking_addr] += amount;
+        } else {
+            validator_state.reward_participators[staking_addr] = amount;
+        }
+    }
+
+    // Use saturating_add for better robustness
+    if reward_state.last_distribution_time + network_parameter.reward_period_seconds
+        > node_state.block_time {
+        // Don't distribute rewards if reward period not reached yet
+        return None;
+    }
+    reward_state.last_distribution_time = node_state.block_time;
+
+    // Mark the dirty flag as soon as reward period reached.
+    node_state.rewards_pool_updated = true;
+  
+    let minted = min(
+      // Use saturating_sub to prevent underflow
+      network_parameter.monetary_expansion_cap - reward_state.minted,
+      // Refer to the following sections for the implementation
+      monetary_expansion(
+          // Refer to the following sections for the implementation
+          total_staking(),
+          reward_state.current_tau,
+          network_parameter.monetary_expansion_r0,
+          network_parameter.reward_period_seconds,
+      )
+    );
+    reward_state.minted += minted;
+
+    // Decay the tau parameter, use the overflow avoid method mentioned above
+    reward_state.current_tau = mul_micro(
+        reward_state.current_tau,
+        network_parameter.monetary_expansion_decay,
+    );
+
+    // Total reward pool for this period
+    let reward_pool = reward_state.period_bonus + minted;
+
+    // Use saturating_add for better robustness
+    let sum_power = validator_state.reward_participators.values().sum();
+
+    // Compute the rewards distribution,
+    // compute `reward_pool * power / sum_power` for each participator
+    let distribution = vec![];
+    if sum_power == 0 {
+        // If there is zero participator, the newly minted coins should be left to the next period
+        reward_state.period_bonus += minted;
+    } else {
+        // Remaining coins after distribution
+        let remains = reward_pool;
+
+        for (addr, power) in validator_state.signed_voters {
+            // Use some type bigger than u64 (e.g. u128) to prevent overflow of multiplication
+            let amount = reward_pool * power / sum_power;
+            remains -= amount;
+            distribution.push((addr, amount));
+        }
+
+        // Remains of distribution is left to the next period
+        reward_state.period_bonus = remains;
+    }
+
+    // Add the rewards to the accounts, implementation is in following sections
+    add_rewards(distribution);
+
+    // Clear rewards statistics
+    validator_state.signed_voters.clear();
+
+    // The Result (distribution, minted) should be written into block result event list
+}
+```
+
+##### `monetary_expansion`
+
+TODO
+
+##### `total_staking`
+
+```rust
+fn total_staking() {
+    let sum = 0;
+    for staking_addr in validator_state.validators {
+        // Panic or not is decided by implementator
+        let account = get_account(staking_addr).unwrap();
+        sum += account.bonded;
+    }
+    sum
+}
+```
+
+##### `add_rewards`
+
+```rust
+// accounts should be stored in merkle trie, the details are ignored here.
+fn add_rewards(dist) {
+    for staking_addr, amount in dist {
+        // Panic or not is decided by implementator
+        let account = get_account(staking_addr).unwrap();
+        account.bonded += amount;
+        account.nonce += 1;
+    }
+}
+```
+
+#### Commit
+
+Update `last_block_height` if `RewardsPoolState` changed:
+
+```rust
+if node_state.reward_state_updated {
+    reward_state.last_block_height = current_block_height;
+    node_state.reward_state_updated = false;
+}
+```
+
+### Hooks
+
+#### When validator becomes inactive at `end_block`
+
+When validator get jailed or unbonded enough to become inactive during a reward period, it's reward participation record should be removed, so it won't receive any rewards.
+
+```rust
+validator_state.signed_voters.remove(staking_addr)
+```
+
+#### Slashing executed
+
+When slashing executed, the `slashed_amount` needs to be added to `period_bonus` of `RewardState`.
+
+```rust
+reward_state.period_bonus += slashed_amount;
+node_state.rewards_pool_updated = true;
+```
+
+#### Fee collected
+
+When transaction fees are collected, the amount needs to be added to `period_bonus` of `RewardState`.
+
+```rust
+reward_state.period_bonus += fees;
+node_state.rewards_pool_updated = true;
+```
+
--- a/docs/modules/rewarding.md
+++ b/docs/modules/rewarding.md
@@ -60,7 +60,7 @@ During the reward period, the following sources of funds are accumulated into re
 - [Transaction Fees](../getting-started/transaction.md#transaction-fee)
 - Slashings
 
-The reward pool get accumulated through the whole reward period, it get distributed once at the end of reward period.
+The reward pool gets accumulated throughout the whole reward period; it gets distributed once at the end of each reward period.
 
 Because of integer calculations, there will be a few remained funds after distribution, they should be left to the next period.
 
@@ -386,4 +386,3 @@ When transaction fees are collected, the amount needs to be added to `period_bon
 reward_state.period_bonus += fees;
 node_state.rewards_pool_updated = true;
 ```
-
--- a/docs/modules/rewarding.md
+++ b/docs/modules/rewarding.md
@@ -62,7 +62,7 @@ During the reward period, the following sources of funds are accumulated into re
 
 The reward pool gets accumulated throughout the whole reward period; it gets distributed once at the end of each reward period.
 
-Because of integer calculations, there will be a few remained funds after distribution, they should be left to the next period.
+Due to fixed point calculations, there may be a few remainders from arithmetic operations after distribution. These remainder amounts should be left in the rewards pool for the next period.
 
 If there is zero participator for the period, the rewards pool should be left to the next period, rather than get burned.
 


devashishdxt (changes requested):
This documentation should be detailed enough for non-developers to understand how rewards work. I guess, it'd be good to provide an overview of everything in "English" (for non-developers) and also in "Pseudo-code" (for developers).

Currently, it just copy-pastes the implementation which is as good as reading code.

yihuang:
> This documentation should be detailed enough for non-developers to understand how rewards work. I guess, it'd be good to provide an overview of everything in "English" (for non-developers) and also in "Pseudo-code" (for developers).
> 
> Currently, it just copy-pastes the implementation which is as good as reading code.

The abstract part should be enough for non-developers?
For the implementation part, I find the rewarding implementation is quite straight forward, the real code is already readable enough for developers I think.

docs/modules/rewarding.md
```diff
@@ -0,0 +1,287 @@
+# Rewarding
+
+## Abstract
+
+Rewarding module enable crypto chain to reward protocol participators, the design should meet following requirements:
+
+- Fixed total supply.
+- Fairness, all honest participators should get rewards, non-participators should not.
+- The rewards amount a validator received should be proportional to the voting power.
```
tomtau: the last point is a bit unclear -- the emitted amount to all validators is proportional to the total vote power. The received share should be proportional only to participation?

docs/modules/rewarding.md
```diff
@@ -0,0 +1,287 @@
...
+
+TODO
+
+## Implementation
+
+### Network parameters
+
+```rust
+pub struct RewardsParameters {
+    /// Time inteval in seconds to do rewards distribution
```
tomtau:
```suggestion
    /// Time interval in seconds to do rewards distribution
```

docs/modules/rewarding.md
```diff
@@ -0,0 +1,287 @@
...
+    pub reward_period_seconds: u64,
+
+    /// Maximum monetary expansion for rewards.
+    pub monetary_expansion_cap: Coin,
+    /// Monetary expansion formula parameter
+    pub monetary_expansion_r0: Milli,
+    /// Monetary expansion formula parameter
+    pub monetary_expansion_tau: u64,
+    /// Monetary expansion formula parameter
+    pub monetary_expansion_decay: u64,
```
tomtau: required ranges?

tomtau: should `monetary_expansion_tau` be called `initial_tau` or something like that if it decays?

docs/modules/rewarding.md
```diff
@@ -0,0 +1,287 @@
...
+    /// Reward pool other than the monetary expansion accumulated in current period
+    pub period_bonus: Coin,
+    /// last block height that updated it (i64 from Tendermint protobuf)
+    pub last_block_height: BlockHeight,
+    /// last reward distribution time, default to the genesis time
+    pub last_distribution_time: Timespec,
+    /// Record how many coins have been minted, can't exceed the cap
+    pub minted: Coin,
+    /// Parameter in monetary expansion formula, decayed for each rewards distribution
+    pub tau: u64,
```
tomtau: `current_tau` to make it clearer that it may be different from the initial one?

docs/modules/rewarding.md
```diff
@@ -0,0 +1,287 @@
...
+### ABCI Events
+
+#### Begin block
+
+```rust
+    fn begin_block(&mut self, req: &RequestBeginBlock) -> ResponseBeginBlock {
+        self.last_state
+            .as_mut()
+            .unwrap()
+            .validators
```
tomtau: this is a sort of "implementation detail" and likely to change

docs/modules/rewarding.md
```diff
@@ -0,0 +1,287 @@
...
+    pub fn record_voters_for_rewarding(
+        &mut self,
+        addrs: impl Iterator<Item = TendermintValidatorAddress>,
+    ) {
+        for addr in addrs {
+            if let Some(staking_address) = self.tendermint_validator_addresses.get(&addr) {
+                if let Some(voting_power) = self
+                    .validator_state_helper
+                    .get_voting_power(staking_address)
+                {
```
tomtau: same

tomtau (changes requested):
the doc should be clear and compact, so one can easily judge whether the code does what it should do -- if there's a copy-pasted super-detailed code (rather than just high level snippets), it'll likely get out of date... and it may hard to judge whether the evolved code still does what it should be doing

docs/modules/rewarding.md
tomtau: many unnecessary details all below (plus the semantics of many of these detailed items isn't define, unless one looks at the code right now, so can be out of date -- e.g. `self.tendermint_validator_addresses`)

docs/modules/rewarding.md
```diff
@@ -0,0 +1,304 @@
+# Rewarding
+
+## Abstract
+
+Rewarding module enable crypto chain to reward protocol participators, the design should meet following requirements:
+
+- Fixed total supply.
+- Fairness, all honest participators should get rewards, non-participators should not.
```
devashishdxt: Define fairness?

docs/modules/rewarding.md
```diff
@@ -0,0 +1,304 @@
+# Rewarding
+
+## Abstract
+
+Rewarding module enable crypto chain to reward protocol participators, the design should meet following requirements:
+
+- Fixed total supply.
```
devashishdxt: Fixed total supply of what?

docs/modules/rewarding.md
```diff
@@ -0,0 +1,304 @@
...
+- Fairness, all honest participators should get rewards, non-participators should not.
+- The rewards amount a validator received should be proportional to the voting power.
+
+The basic idea is to record signing voters for each block, along with the voting power at the vote time. Distribute the reward pool according to the voting powers.
```
devashishdxt: Describe more with an example?

docs/modules/rewarding.md
```diff
@@ -0,0 +1,304 @@
...
+#### Fixed point arithemetic
+
+TODO
+
+## Implementation
+
+### Network parameters
+
+```rust
+pub struct RewardsParameters {
```
devashishdxt, line 134: At the start of the document, can you explain the role each of these network parameters and how they change reward behavior?

docs/modules/rewarding.md
```diff
@@ -0,0 +1,304 @@
...
+
+The basic idea is to record signing voters for each block, along with the voting power at the vote time. Distribute the reward pool according to the voting powers.
+
+### Concepts
+
+#### Signing voters
+
+Signing voters are defined as:
+
+```rust
```
devashishdxt, line 44: One needs to know **Rust** and **ABCI internals** to understand what are **Signing voters**. Maybe add a bit more description?

docs/modules/rewarding.md
```diff
@@ -0,0 +1,304 @@
...
+  .last_commit_info
+  .votes
+  .filter(|vote| vote.signed_last_block)
+```
+
+All the signing voters of each block get rewarded, if a validator don't participate in voting, it doesn't get rewarded.
+
+#### Reward period
+
+Rewards are distributed to the validators periodically.
```
devashishdxt: Also specify that this can be changed using network parameter in `genesis.json`.

docs/modules/rewarding.md
```diff
@@ -0,0 +1,304 @@
...
+
+#### Reward pool
+
+During the reward period, the following sources of rewarding are accumulated into reward pool:
+
+- Monetary expansion
```
devashishdxt: What does monetary expansion mean? Add a small definition and also link to the **Monetary expansion** section.

docs/modules/rewarding.md
```diff
@@ -0,0 +1,304 @@
...
+}
+
+pub struct ChainNodeApp {
+    pub rewards_pool_updated: bool
+}
+```
+
+### ABCI Events
+
+#### Begin block
```
devashishdxt, line 221: Maybe specify **_in English_**, what should happen during `BeginBlock` and don't expose the whole implementation (as it may change in future).

docs/modules/rewarding.md
```diff
@@ -0,0 +1,304 @@
...
+```rust
+    pub fn commit_handler(&mut self, _req: &RequestCommit) -> ResponseCommit {
+        if self.rewards_pool_updated {
+            top_level.rewards_pool.last_block_height = new_state.last_block_height;
+            self.rewards_pool_updated = false;
+        }
+    }
+```
+
+### Hooks
```
devashishdxt, line 362: No explanation of these **Hooks** and also no mention of where these should be called from.

docs/modules/rewarding.md
```diff
@@ -0,0 +1,304 @@
...
+
+New coins are minted into the reward pool at the end of reward period. The amount is calculated based on the total staking of the last block:
+
+```
+R0 = rewards_config["monetary_expansion_r0"]
+tau = rewards_config["monetary_expansion_tau"]
+P = rewards_config["distribution_period"]
+
+# total bonded amount of the active validators at the end of the reward epoch
+S = total_staking
```
devashishdxt: How is `total_staking` calculated?

devashishdxt (changes requested):
There are still a few missing pieces which this spec does not explain:
- What happens when a validator gets jailed in a reward period?
- What happens when a validator unbonds their funds in a reward period?

Also, it is not clear what should happen in each ABCI call `InitChain`, `BeginBlock`, `EndBlock`, etc. One has to read and understand the whole code to know the behavior of rewards module.

I think it is better to merge this with [Validator Rewards](https://github.com/crypto-com/chain-docs/blob/master/docs/getting-started/staking.md#validator-rewards) section in main documentation (move everything to a new file?) which already explains everything in good detail.

docs/modules/rewarding.md
yihuang: I changed to "The rewards amount distributed to a validator should be proportional to his voting power", the minted amount will be explained in the monetary expansion section, it's related to total staking, but not necessarily proportional.

docs/modules/rewarding.md
yihuang: > required ranges?

Added the validation code.

yihuang: > should `monetary_expansion_tau` be called `initial_tau` or something like that if it decays?

Since they are all in `InitNetworkParameters`, I think the init part if implicit already?

docs/modules/rewarding.md
yihuang: Done

docs/modules/rewarding.md
yihuang: The sentence after it explains it: "all honest participators should get rewards, non-participators should not".

docs/modules/rewarding.md
yihuang: Changed it to "The total supply of CRO is fixed"

docs/modules/rewarding.md
yihuang: "All the active validators who follow the protocol honestly will be recorded as signing voters."
Added this sentence, should be enough for people to understand it.

docs/modules/rewarding.md
yihuang: Done

docs/modules/rewarding.md
yihuang: Added a link.

yihuang:
Just cleaned up the code, removed the accidental things into separated functions.
Now there's very little thing that can be removed without affecting the logic.

docs/modules/rewarding.md
yihuang: There is a "total staking" section now which explains it.

docs/modules/rewarding.md
yihuang: Added more descriptions in the hook section, please review to see if it's clear enough.

docs/modules/rewarding.md
yihuang: Added some simple descriptions before that code.

docs/modules/rewarding.md
yihuang: I think the parameters are all explained in different places in part 1.

yihuang:
> There are still a few missing pieces which this spec does not explain:
> 
> * What happens when a validator gets jailed in a reward period?
> * What happens when a validator unbonds their funds in a reward period?

They are explained in the section hooks, added more descriptions.

> 
> Also, it is not clear what should happen in each ABCI call `InitChain`, `BeginBlock`, `EndBlock`, etc. One has to read and understand the whole code to know the behavior of rewards module.
> 

The code is cleaned up now, with comments, it should be quite readable to a developer.

> I think it is better to merge this with [Validator Rewards](https://github.com/crypto-com/chain-docs/blob/master/docs/getting-started/staking.md#validator-rewards) section in main documentation (move everything to a new file?) which already explains everything in good detail.

docs/modules/rewarding.md
yihuang: Cleaned up the code, moved the accidental things into separated functions.

docs/modules/rewarding.md
yihuang: Cleaned up the code, moved the accidental things into separated functions.

yihuang:
> the doc should be clear and compact, so one can easily judge whether the code does what it should do -- if there's a copy-pasted super-detailed code (rather than just high level snippets), it'll likely get out of date... and it may hard to judge whether the evolved code still does what it should be doing

I think with the cleaned-up version of code, there is very little that is not directly related to the logic. And it's the more precise way to describe the logic.

docs/modules/rewarding.md
tomtau: where does this definition come from? 
the common one is:
```
3 (fairness). A process gets a reward for a block if and only if it has a reward parameter for that block equal to 1. 
3' (eventual fairness). There exists a height H such that for a block in the blockchain at height H′ > H a process gets a reward for that block if and only if it has a reward parameter for that block equal to 1.
```

docs/modules/rewarding.md
```diff
@@ -0,0 +1,334 @@
...
+
+            for (addr, &power) in self.validator_state().signed_voters.iter() {
+                // Use u128 to prevent overflow of intermidiate results
+                let amount: u128 =
+                    u128::from(reward_pool) * u128::from(power) / u128::from(sum_power);
+                let amount = Coin::new(amount.try_into().unwrap()).unwrap();
+                remains = remains.saturating_sub(amount);
+                distribution.push((*addr, amount));
+            }
+            remains
```
tomtau: this is still "unfair" (as per the agreed definition):

1. it only tracks "IF" the validator correctly participated, not how often
2. it distributes per vote power share, not per participation.

--
(eventually) fair reward distribution should reward a node for each block if and only if it participated correctly (live, non-byzantine).

...
Here, the counter-example is that if there are more than 1 block during that period, it won't be fair. e.g. one node behaves like this:
block 1 -- reward parameter=1
block 2 -- reward parameter=0
block 3 -- reward parameter=0
...
blocks N -> distribution -- gets rewarded for all blocks up to N, regardless how it behaved in each of them which is unfair.

docs/modules/rewarding.md
```diff
@@ -0,0 +1,334 @@
+# Rewarding
+
+## Abstract
+
+Rewarding module enable crypto chain to reward protocol participators, the design should meet following requirements:
+
+- The total supply of CRO is fixed.
+- Fairness, all honest participators should get rewards, non-participators should not.
+- The rewards amount distributed to a validator should be proportional to his voting power.
```
tomtau: The total reward emitted should be proportional to the total voting power,

but the distributed amount should be proportional to the correct participation (assuming distribution doesn't happen every block due to performance degrade).

tomtau (changes requested):
looks a bit simpler, but again the code snippets should ideally omit details that may change or explain them more in detail.
e.g. `self.rewards_pool_updated` is just some temporary flag (the implementation may change) that helps with app hash re-calculations or what not?

docs/modules/rewarding.md
yihuang: total reward emitted could be positively or negatively related to the total voting power, not necessarily proportional(maybe I have a wrong understanding of proportional?), also it could be unrelated to total staking at all if monetary expansion cap is reached.
"correct participation" is the second point?

yihuang:
> looks a bit simpler, but again the code snippets should ideally omit details that may change or explain them more in detail.
> e.g. `self.rewards_pool_updated` is just some temporary flag (the implementation may change) that helps with app hash re-calculations or what not?

`self.rewards_pool_updated` is part of the logic of how `last_block_height` gets updated, I don't see a way to explain it without this flag.

docs/modules/rewarding.md
yihuang: You seem to try to distribute rewards per block, but I think they are actually equivalent:
```
block 1 -- reward parameter=1, power of A=p, power of others=p'
block 2 -- reward parameter=0
block 3 -- reward parameter=0
...
blocks N -> distribution
``` 
Assuming all the other validators have reward parameter=1, then the reward distribution of A is:

```
total_power = p + p' * N

# current computation
reward_a = reward_pool * p / total_power

# only focus on block 1
reward_block1 = reward_pool * (p + p') / total_power
reward_a = reward_block1 * p / (p + p')
         = reward_pool * p / total_power
```

yihuang: > where does this definition come from?
> the common one is:
> 
> 3 (fairness). A process gets a reward for a block if and only if it has a reward parameter for that block equal to 1. 
> 3' (eventual fairness). There exists a height H such that for a block in the blockchain at height H′ > H a process gets a reward for that block if and only if it has a reward parameter for that block equal to 1.

I think it's just the 3 in layman's term?

docs/modules/rewarding.md
devashishdxt: Where is part 1? Am I missing something?

yihuang: > Where is part 1? Am I missing something?

I mean the section "Abstract".
For example, the monetary expansion parameters R0/tau are in the monetary expansion formula.
The reward period has a section in concepts. I should add more description for the `rewards_monetary_expansion_cap` and `monetary_expansion_decay `.

docs/modules/rewarding.md
yihuang: Added an example now.

docs/modules/rewarding.md
yihuang: Done

docs/modules/rewarding.md
tomtau: normal fairness gets that reward for that block.

eventual fairness gets reward for that block in some time in the future -- the important bit here there is " if and only if it ".

In the current mechanism, yes, it gets a reward for that block AND it also gets rewards for blocks where its reward parameter=0 (as the current mechanism doesn't distinguish correct participation) which it shouldn't

docs/modules/rewarding.md
tomtau: I'm confused if the third point talks about the share or the total emitted reward?

docs/modules/rewarding.md
yihuang: > normal fairness gets that reward for that block.
> 
> eventual fairness gets reward for that block in some time in the future -- the important bit here there is " if and only if it ".
> 
> In the current mechanism, yes, it gets a reward for that block AND it also gets rewards for blocks where its reward parameter=0 (as the current mechanism doesn't distinguish correct participation) which it shouldn't

As I showed in the previous comment, it only gets the reward for blocks where its reward parameter = 1, not get rewards for blocks where its reward parameter=0.
I copy it here:

> You seem to try to distribute rewards per block, but I think they are actually equivalent:
> ```
> block 1 -- reward parameter=1, power of A=p, power of others=p'
> block 2 -- reward parameter=0
> block 3 -- reward parameter=0
> ...
> blocks N -> distribution
> ``` 
> Assuming all the other validators have reward parameter=1, then the reward distribution of A is:
> 
> ```
> total_power = p + p' * N
> 
> # current computation
> reward_a = reward_pool * p / total_power
> 
> # only focus on block 1
> reward_block1 = reward_pool * (p + p') / total_power
> reward_a = reward_block1 * p / (p + p')
>          = reward_pool * p / total_power
> ```
> 
>

docs/modules/rewarding.md
yihuang: > I'm confused if the third point talks about the share or the total emitted reward?

It says "The rewards amount distributed to a validator", so it's the share of a single validator, how should I make it clearer?

docs/modules/rewarding.md
yihuang: Do you mean don't accumulate rewards into a pool for the whole period, but calculate reward distribution for each block separately. Because different blocks might have different fees collected?

docs/modules/rewarding.md
tomtau: Well, one of the points of this was:

1. total emitted amount should be proportional to the total vote power (as before)
2. individual shares should correspond to correct behaviour participation, *regardless* of the vote power

And the correct behaviour is defined with the respect to individual blocks with "IF AND ONLY IF" for each block -- it's not defined as "As long as a validator exhibited correct behaviour in 1 block out of N in a given, he should get reward for all N blocks"

docs/modules/rewarding.md
tomtau: it's defined with "IF AND ONLY IF" -- e.g. two validators A B, both have the same vote power.
Node A:
```
block 1 -- reward parameter=1
block 2 -- reward parameter=0
block 3 -- reward parameter=0
...
```
Node B:
```
block 1 -- reward parameter=1
block 2 -- reward parameter=1
block 3 -- reward parameter=1
...
```

reward gets distributed at N -- currently, the share calculation takes into account only vote power... so here, A and B have the same vote power, so they get rewarded the same.
When interpreted per block, this does not satisfy (eventual) fairness, as participants should be reward IF AND ONLY IF... here, A was rewarded for blocks 2, 3...

docs/modules/rewarding.md
yihuang: “total emitted amount should be proportional to the total vote power” is not accurate, fees and slashings are not related to total vote power, the amount of new minted coins is a function of total vote power, but not necessarily proportional to that.

By "regardless of the vote power", do you mean one vote of M power and one vote of N power should get the same amount of reward?

"he should get reward for all N blocks", I don't understand, in current computation, he indeed only get the rewards for the blocks he participated correctly.

docs/modules/rewarding.md
yihuang: > it's defined with "IF AND ONLY IF" -- e.g. two validators A B, both have the same vote power.
> Node A:
> 
> ```
> block 1 -- reward parameter=1
> block 2 -- reward parameter=0
> block 3 -- reward parameter=0
> ...
> ```
> 
> Node B:
> 
> ```
> block 1 -- reward parameter=1
> block 2 -- reward parameter=1
> block 3 -- reward parameter=1
> ...
> ```
> 
> reward gets distributed at N -- currently, the share calculation takes into account only vote power... so here, A and B have the same vote power, so they get rewarded the same.
> When interpreted per block, this does not satisfy (eventual) fairness, as participants should be reward IF AND ONLY IF... here, A was rewarded for blocks 2, 3...

In this case, the computation is like this:
```
pa = power of A
pb = power of B
sum_power = pa + pb * N
sum_power_a = pa
sum_power_b = pb * N
reward_a = reward_pool * sum_power_a / sum_power
reward_b = reward_pool * sum_power_b / sum_power
```

if interpret it per block:
```
power_block1 = pa + pb
sum_power = pa + pb * N
reward_block1 = reward_pool * power_block1 / sum_power
                = reward_pool * (pa + pb) / sum_power
reward_a = reward_block1 * pa / (pa + pb)
           = reward_pool * pa / sum_power
```

The result is the same.

docs/modules/rewarding.md
tomtau: > “total emitted amount should be proportional to the total vote power” is not accurate, fees and slashings are not related to total vote power, the amount of new minted coins is a function of total vote power, but not necessarily proportional to that.

emitted means it's newly emitted, so doesn't include fees or slashing -- perhaps more precise is inversely proportional or what's out of the formula.
it can say that the "total reward includes total emitted coins (proportional to ...) and bonus"

> By "regardless of the vote power", do you mean one vote of M power and one vote of N power should get the same amount of reward?

It depends on their correct behaviour.
If both behaved correctly (live, non-byzantine) at all times, yes

docs/modules/rewarding.md
tomtau: Oh, https://github.com/crypto-com/chain-docs/pull/94/files#diff-2c819a76ed9cc68e5c8a8d51dc8ac205R208 there's a `saturating_add`

tomtau: I skimmed through it and only saw "insert"

docs/modules/rewarding.md
yihuang: > > By "regardless of the vote power", do you mean one vote of M power and one vote of N power should get the same amount of reward?
> 
> It depends on their correct behaviour.
> If both behaved correctly (live, non-byzantine) at all times, yes

Do you mean a validator with 10 power get the same amount of rewards as a validator with 1 power, if they both behaved correctly at all times? Isn't this incentivizes the validator with more power to separate his power into multiple smaller validators, and takes more validator spots, and which is something we want to avoid?

docs/modules/rewarding.md
tomtau: Anyway, this shows that a more high-level concise description or pseudocode will be clearer than 100+ lines of detailed code where it is hard to spot the important parts

tomtau: `record_voters_for_rewarding` could just be something like:
```
if signed_voters[staked_address]:
signed_voters[staked_address] += vote_power_at_X
else:
signed_voters[staked_address] = vote_power_at_X
```

yihuang: > `record_voters_for_rewarding` could just be something like:
> 
> ```
> if signed_voters[staked_address]:
> signed_voters[staked_address] += vote_power_at_X
> else:
> signed_voters[staked_address] = vote_power_at_X
> ```

But it will miss two informations:
- overflow behavior
- Error handling when address and voting power not found

devashishdxt (commented):
A few thoughts:

This document exposes a lot of implementation details (it exposes the whole implementation). Specifications, generally, are supposed to act as a blueprint to help developers implement the functionality. It should describe all the flows in the functionality and also specify behavior of the program for each flow. Besides this, it is ok to provide high level interfaces that can be used by developer but, detailed implementation shouldn't be a part of specification.

Understanding the behaviour of rewards module from this document requires going through the code and understanding it, which is not what specification is meant for. Ideally, `specification + developer = code`. Whereas here, `specification + code + developer = more code`.

Any thoughts?

yihuang:
> A few thoughts:
> 
> This document exposes a lot of implementation details (it exposes the whole implementation). Specifications, generally, are supposed to act as a blueprint to help developers implement the functionality. It should describe all the flows in the functionality and also specify behavior of the program for each flow. Besides this, it is ok to provide high level interfaces that can be used by developer but, detailed implementation shouldn't be a part of specification.
> 
> Understanding the behaviour of rewards module from this document requires going through the code and understanding it, which is not what specification is meant for. Ideally, `specification + developer = code`. Whereas here, `specification + code + developer = more code`.
> 
> Any thoughts?

I guess it all come down to how precise do we want the specification to be, or what functionality do we want from the specification, do we want it to be some descriptive documentation, or precise stuff which can be used to check the correctness.
For example, the error handling(panic or not in some edge cases) or overflow behavior(panic, wrap, saturate) is important part of the correctness, many seemingly verboseness in the code are  actually these things.

devashishdxt:
> > A few thoughts:
> > This document exposes a lot of implementation details (it exposes the whole implementation). Specifications, generally, are supposed to act as a blueprint to help developers implement the functionality. It should describe all the flows in the functionality and also specify behavior of the program for each flow. Besides this, it is ok to provide high level interfaces that can be used by developer but, detailed implementation shouldn't be a part of specification.
> > Understanding the behaviour of rewards module from this document requires going through the code and understanding it, which is not what specification is meant for. Ideally, `specification + developer = code`. Whereas here, `specification + code + developer = more code`.
> > Any thoughts?
> 
> I guess it all come down to how precise do we want the specification to be, or what functionality do we want from the specification, do we want it to be some descriptive documentation, or precise stuff which can be used to check the correctness.
> For example, the error handling(panic or not in some edge cases) or overflow behavior(panic, wrap, saturate) is important part of the correctness, many seemingly verboseness in the code are actually these things.

These are important parts of a program's implementation and most of the specifications don't delve into these implementation details because the idea to to keep them agnostic of how any developer chooses to implement them (for example, decisions like which data sctucture to use, what language to use, how to handle errors, etc. are left to developers). Here is an example, [this](https://www.ietf.org/rfc/rfc793.txt) is **Transmission Control Protocol** specification and they explain the whole protocol without code.

P.S.: I'm ok with writing the whole code in specification if everyone is comfortable with this.

docs/modules/rewarding.md
tomtau: Ignore that -- it was my misunderstanding of how the activity is recorded.

docs/modules/rewarding.md
tomtau: Is any of them expected to happen as a part of normal behaviour?

It'll ideally list all its assumptions / invariants separately -- e.g. it's assumed that one can lookup staking address (based on the consensus pubkey/address) for the duration of the reward period, unless it's been removed due to punishment mechanisms?

For expected behaviour, it should specify these.
For unexpected behaviour, there should be an overall note around handling -- e.g. unexpected behaviour / invariant breaking behaviour will be logged with "error" level in deployment builds, but will panic in debug/fuzzing/test builds?

tomtau (changes requested):
the revisions look better -- it'll be more comprehensive if the code snippets are shorter and only capture the important part.
also some details may be missing ? e.g. `add_rewards` isn't documented -- nonce is increased by 1? something else?

Everything that's not common Rust, Tendermint ABCI or defined elsewhere in chain-docs is subject to change (all these `self.<something>` ...), so it's better to be omitted and rephrased in more future-proof way -- a sort of: here are the inputs, here are assumptions / invariants about these inputs, this is how they are processed, these are the outputs.

docs/modules/rewarding.md
```diff
@@ -0,0 +1,389 @@
...
+
+#### Reward pool
+
+During the reward period, the following sources of funds are accumulated into reward pool:
+
+- [Monetary expansion](#monetary-expansion)
+- [Transaction Fees](../getting-started/transaction.md#transaction-fee)
+- Slashings
+
+The reward pool get accumulated through the whole reward period, it get distributed once at the end of reward period.
```
tomtau:
```suggestion
The reward pool gets accumulated throughout the whole reward period; it gets distributed once at the end of each reward period.
```

docs/modules/rewarding.md
```diff
@@ -0,0 +1,389 @@
...
+
+Because of integer calculations, there will be a few remained funds after distribution, they should be left to the next period.
```
tomtau:
```suggestion
Due to fixed point calculations, there may be a few remainders from arithmetic operations after distribution. These remainder amounts should be left in the rewards pool for the next period.
```

docs/modules/rewarding.md
```diff
@@ -0,0 +1,389 @@
...
+
+If there is zero participator for the period, the rewards pool should be left to the next period, rather than get burned.
```
tomtau, line 67: does this mean that if there were no participants, reward distribution should not happen? When would this happen?

docs/modules/rewarding.md
```diff
@@ -0,0 +1,389 @@
...
+The main actions all happens in begin block event, it's the time we get reports of participators through `last_commit_info`, also the time the node state is consistent.
+
+1. Collect the validator addresses of reward participators from `last_commit_info`.
+2. Record them, and accumulate their voting powers.
+3. If a reward period has passed, distribute the reward pool according to the participator statistics.
+
+##### `begin_block`
+
+```rust
+fn begin_block(req: RequestBeginBlock) {
```
tomtau, line 232: just side-note that I guess it should happen after punishment-related begin block processing? there's a separate issue for this / how to generate validator updates: https://github.com/crypto-com/chain-docs/issues/81 so no need in this PR

tomtau (changes requested):
looks all right and much clearer -- left a few small suggestions, but otherwise good to go

docs/modules/rewarding.md
yihuang: Maybe, all the previous participators get jailed in node state somehow, but still manage to produce the last block of reward period, because tendermint validator set don't updated yet.

PR #97, opened 2020-03-03
LivenessTracker should be updated in EndBlock

devashishdxt:
Changes:

- `LivenessTracker` should be updated in `EndBlock`. Reason: If we do not update `LivenessTracker` in `EndBlock`, we'll have to put update statements in all the places where a validator's voting power changes to zero (jailing, unbonding, etc.). On the other hand, if we update `LivenessTracker` in `EndBlock` based on `power_changed_in_block`, we won't have to worry about updating liveness tracker in different places.
- Added instruction to add a new validator to `LivenessTracker` when they join the network as a validator using `NodeJoinTx`.
- Fixed some typos.

### DIFF
--- a/docs/getting-started/punishments.md
+++ b/docs/getting-started/punishments.md
@@ -233,10 +233,6 @@ impl LivenessTracker {
     - Add validator address to a list of permanently banned validators.
     - Generate slashing and jailing events for validator.
 
-#### `DeliverTx`
-
-- If a new node joins the validator set using `NodeJoinTx`, add the validator address of that node to `LivenessTracker`.
-
 :::tip Note:
 An additional validation for `NodeJoinTx` will be to verify if the validator address is not present in the list of
 permanently banned validators.
@@ -244,6 +240,7 @@ permanently banned validators.
 
 #### `EndBlock`
 
-- Remove all the validators from `LivenessTracker` whose voting power was changed to zero in the block.
+- Update `LivenessTracker`, i.e., add all the validators who joined in current block and remove all the validator who,
+  left/got jailed in current block.
 - Set `response.validator_updates` for all the validators whose voting power was changed. Tendermint will remove all the
   validators whose voting power is set to zero, so, all the jailed validators should have zero voting power.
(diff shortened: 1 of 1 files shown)


docs/getting-started/punishments.md
```diff
@@ -225,17 +228,23 @@ impl LivenessTracker {
...
+
 :::tip Note:
 An additional validation for `NodeJoinTx` will be to verify if the validator address is not present in the list of
 permanently banned validators.
 :::
 
 #### `EndBlock`
 
+- Remove all the validators from `LivenessTracker` whose voting power was changed to zero in the block (data can be
+  obtained from `power_changed_in_block`).
```
tomtau: `power_changed_in_block` is undefined -- some WIP of how this can be retrieved was described in: https://github.com/crypto-com/chain-docs/pull/75/files

devashishdxt: Added an overview of how validator changes work in code and also explained `power_changed_in_block`. Please check.

docs/getting-started/punishments.md
```diff
@@ -228,11 +228,21 @@ impl LivenessTracker {
...
     - Generate slashing and jailing events for validator.
 
+:::tip How are validator updates handled in code?
+Tendermint expects all the validator updates to be done in `EndBlock`, i.e., all the validator updates are posted to
+tendermint in `EndBlock`'s response. So, we need a mechanism to store all the validator updates happened in a block and
+post those to tendermint in `EndBlock`'s response. For this purpose, we have `power_changed_in_block` (implementation
+and variable name may vary but the idea is the same) where we store all the changes in voting power that happened in a
+block. `power_changed_in_block` is cleared at the start of each new block.
+
+So, setting voting power in `power_changed_in_block` to zero means that those validators will be removed.
```
tomtau: Not necessarily -- e.g. in one block
"unbondtx (some amount) -> sets power to 0"
"deposit (other amount) -> sets power back up"

depending on the amount, and other actions in that block, one can't simply tell what will happen in end block.

`power_changed_in_block` was just this temporary internal helper that's been regretfully misused and shouldn't be relied on as some sort of universal validator update semantics (that's not defined anywhere).
Punishments may require setting more things than just that.

The only important part here is what happens with regards to ABCI -- which is that at end block, it should generate a ValidatorUpdate for the node's associated consensus pub key and set the vote power to 0.

tomtau (changes requested):
the note is a bit misleading.
what should be there is perhaps blocked by: https://github.com/crypto-com/chain-docs/issues/81 ?

devashishdxt:
> the note is a bit misleading.
> what should be there is perhaps blocked by: #81 ?

I agree with you. Modified the PR (removed all the mentions of `power_changed_in_block`).

PR #102, opened 2020-03-13
Problem (Fix #81): reward<->punishment<->monetary update interactions not explicitly stated

### LINKED ISSUE #81: Problem: reward<->punishment<->monetary update interactions not explicitly stated

yihuang:
Solution:
- Add a doc to describe staking/validator state transitions.

### DIFF
--- a/docs/modules/staking-state.md
+++ b/docs/modules/staking-state.md
@@ -1,0 +1,210 @@
+# Staking state and transitions
+
+```rust
+struct StakedState {
+    address: StakedStateAddress,
+    nonce: u64,
+    bonded: Coin,
+    unbonded: Coin,
+    unbonded_from: Timespec,
+    validator: Option<Validator>,
+}
+
+struct Validator {
+    council_node: CouncilNode,
+    jailed_until: Option<Timespec>,
+    inactive_time: Option<Timespec>,
+    old_validator_keys: Vec<TendermintValidatorPubKey>,
+}
+```
+
+## States
+
+### Clean staking
+
+`validator.is_none()`
+
+### Validator
+
+`validator.is_some()`
+
+There are several variants of it:
+
+#### Active
+
+`validator.inactive_time.is_none()`
+
+> **_NOTE:_** Active validator doesn't nessesarily mean the final validator take effect in tendermint, please refer to
+> [Choose final validators](#choose-final-validators)
+
+#### Inactive
+
+`validator.inactive_time.is_some()`
+
+#### Jailed
+
+`validator.jailed_until.is_some()`
+
+> **_NOTE:_** Jailed implies inactive, but not vice versa
+
+## State transitions
+
+### From clean staking or inactive validator to active validator
+
+#### Node join
+
+The only way to transit to active validator is by executing `NodeJoinTx`, the preconditions is:
+
+- `bonded >= min_required_staking`
+- The validator pubkey/address is not banned or already used
+- Not jailed if transitting from inactive validator
+
+### From active validator to inactive validator
+
+There are several cases for this:
+
+#### Bonded coins become lower than required
+
+When `bonded < min_required_staking`, this transition happens.
+
+The reasons for dropping of bonded coins are:
+
+- Execute `UnbondTx` at `deliver_tx` event
+- Slashed for non-live or byzantine faults at `begin_block` event
+
+> **_NOTE:_** The transition happens immediatelly in `deliver_tx`, won't reverse automatically when bonded coins become
+> enough again even in the same block
+
+#### Jailed for byzantine faults
+
+Jailed always implies inactive.
+
+This happens in `begin_block` event.
+
+### From inactive validator to clean staking
+
+#### Clean up
+
+The clean up procedure will remove the validator record if:
+
+- Not jailed
+- `block_time >= inactive_time + cleanup_period`
+
+> _**NOTE:**_ `cleanup_period`
+>
+> Currently `cleanup_period = unbonding_period`, but logically, `cleanup_period` only needs such constraints:
+>
+> - `> max_evidence_age`, so we can handle delayed byzantine evidences (inactive validator can still be slashed for
+>   later detected byzantine faults)
+> - `> 2 blocks`, so when we don't panic when seeing signing vote of inactivated validators
+
+### From jailed validator to inactive(unjailed) validator
+
+#### Unjail
+
+The only way to leave jailed validator state is by executing `UnjailTx`, the preconditions are:
+
+- Already jailed
+- `block_time >= jailed_until`
+
+## Appendix
+
+### Choose final validators
+
+The final validator set that take effect in tendermint is chosen at `end_block` event by:
+
+- Sort all the active validators by `voting_power desc, staking_address`
+- Take the first `max_validators` ones
+
+For example, if you are the `max_validators + 1`th active validator, so you are not chosen yet, but if anyone in the
+chosen set quit, you will be chosen automatically at the next `end_block` event.
+
+### Implication of jailing
+
+#### Transactions
+
+Only `UnjailTx` is allowed to be executed on a jailed staking if the `jailed_until` time is passed.
+
+Disallowed transactions are:
+
+- `DepositTx`
+- `WithdrawTx`
+- `UnbondTx`
+- `NodeJoinTx`
+
+#### Reward distribution
+
+It won't distribute rewards to jailed validators, but inactive and not jailed validators get rewards as normal.
+
+When a validator is jailed, it's reward participation tracking records is removed immediatelly.
+
+### Nonce
+
+The `nonce` is increased once and exactly once for each operations mutate the `StakedState`:
+
+- All the transaction types that mutate staking state:
+
+  - `DepositTx`
+  - `WithdrawTx`
+  - `UnbondTx`
+  - `UnjailTx`
+  - `NodeJoinTx`
+
+- Reward distribution
+
+- Slashing for non-live or byzantine faults
+
+  > _**NOTE:**_ Slashings for different faults are considered independent operations, each slashing will increase nonce independently.
+
+- [Clean up](#clean-up)
+
+### Liveness tracking
+
+- All active validators's liveness trackers are maintained no matter if it's chosen into the [final validator set](#choose-final-validators).
+
+  If it doesn't appear in the votes reported by tendermint, it's recorded as a `true` which means live by default.
+
+- Inactive validator's liveness trackers are also maintained, this serve two purposes:
+
+  - After a validator invalidated, the signing vote might still arrive for the next two blocks, we don't want to issue a
+    false warning in this case.
+  - Validator might quit and re-join very fast (by `UnbondTx`/`DepositTx`/`NodeJoinTx`), in this case it's liveness
+    tracking record will preserve.
+
+So, the liveness tracker is only removed when validator record get [cleaned up](#clean-up).
+
+### Inactive validator re-join with different validator key
+
+When an inactive validator re-join, it can provide different validator key, but it still need to be hold responsible for
+byzantine fault committed before for as long as `max_evidence_age`. So we need to keep the old validator pub keys for
+sometime.
+
+Currently, it's always kept in `old_validator_keys` before the whole validator record get [cleaned up](#clean-up). These
+keys can't be re-used by other validators before cleaned up.
+
+### Validator key blacklist
+
+For all the byzantine faults we received, as long as the evidene is valid, the validator key is put into a blacklist,
+it's banned permanently.
+
+### Non exists and empty staking 
+
+Empty staking is defined as:
+
+```rust
+StakedState {
+    address,
+    nonce: 0,
+    bonded: 0,
+    unbonded: 0,
+    unbonded_from: 0,
+    validator: None,
+}
+```
+
+For all the logic processing, the result of success execution should be the same for non exists staking and empty staking.
+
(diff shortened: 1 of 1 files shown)


docs/getting-started/punishments.md
```diff
@@ -89,13 +91,15 @@ When a jailed validator wishes to resume normal operations (after `account.jaile
...
+calculated based on `BYZANTINE_SLASH_PERCENT`.
 
 :::warning Important:
-A validator should not be slashed more than once for byzantine faults within `UNBONDING_PERIOD`. If a validator commits multiple byzantine faults
-within that time period, it should only be slashed once (for simplicity, we'll only slash the validator for the first
-evidence that we get from tendermint and ignore other evidences until `UNBONDING_PERIOD`).
-:::
+A validator should not be slashed more than once for byzantine faults within `UNBONDING_PERIOD`. If a validator commits
+multiple byzantine faults within that time period, it should only be slashed once (for simplicity, we'll only slash the
+validator for the firs evidence that we get from tendermint and ignore other evidences until `UNBONDING_PERIOD`).  :::
```
lezzokafka: typo?

tomtau: firs -> first?

yihuang: Oh, sorry, the punishments are added by mistake, just removed it from the commit.

docs/modules/staking-state.md
```diff
@@ -0,0 +1,210 @@
...
+    unbonded: Coin,
+    unbonded_from: Timespec,
+    validator: Option<Validator>,
+}
+
+struct Validator {
+    council_node: CouncilNode,
+    jailed_until: Option<Timespec>,
+    inactive_time: Option<Timespec>,
+    old_validator_keys: Vec<TendermintValidatorPubKey>,
```
tomtau: Why to keep `old_validator_keys` in here?

docs/modules/staking-state.md
```diff
@@ -0,0 +1,210 @@
...
+
+## State transitions
+
+### From clean staking or inactive validator to active validator
+
+#### Node join
+
+The only way to transit to active validator is by executing `NodeJoinTx`, the preconditions is:
+
+- `bonded >= min_required_staking`
```
lezzokafka, line 58: Align with [EFFECTIVE_MIN_STAKE](https://crypto-com.github.io/getting-started/staking.html#effective-min-stake)?

docs/modules/staking-state.md
```diff
@@ -0,0 +1,210 @@
...
+- The validator pubkey/address is not banned or already used
+- Not jailed if transitting from inactive validator
+
+### From active validator to inactive validator
+
+There are several cases for this:
+
+#### Bonded coins become lower than required
+
+When `bonded < min_required_staking`, this transition happens.
```
lezzokafka, line 68: Same as the  line 58 comment

docs/modules/staking-state.md
```diff
@@ -0,0 +1,210 @@
...
+  - `WithdrawTx`
+  - `UnbondTx`
+  - `UnjailTx`
+  - `NodeJoinTx`
+
+- Reward distribution
+
+- Slashing for non-live or byzantine faults
+
+  > _**NOTE:**_ Slashings for different faults are considered independent operations, each slashing will increase nonce independently.
```
lezzokafka, line 197:
```suggestion
  > _**NOTE:**_ Slashings for different faults are considered to be independent operations, each slashing will increase nonce independently.
```

docs/modules/staking-state.md
```diff
@@ -0,0 +1,210 @@
...
+So, the liveness tracker is only removed when validator record get [cleaned up](#clean-up).
+
+### Inactive validator re-join with different validator key
+
+When an inactive validator re-join, it can provide different validator key, but it still need to be hold responsible for
+byzantine fault committed before for as long as `max_evidence_age`. So we need to keep the old validator pub keys for
+sometime.
+
+Currently, it's always kept in `old_validator_keys` before the whole validator record get [cleaned up](#clean-up). These
+keys can't be re-used by other validators before cleaned up.
```
tomtau: what about just blacklisting them (and the blacklist holds pub key / tm address->staking address mapping)? that would enforce the logical many (consensus pub key)-to-one (staking address) rather than many-to-many

docs/modules/staking-state.md
```diff
@@ -0,0 +1,211 @@
...
+
+## State transitions
+
+### From clean staking or inactive(unjailed) validator to active validator
+
+#### Node join
+
+The only way to transit to active validator is by executing `NodeJoinTx`, the preconditions is:
+
+- `bonded >= min_required_staking`
```
tomtau, line 58: so, there's no more "min_effective"? I guess that should then be removed from elsewhere in the docs?

docs/modules/staking-state.md
```diff
@@ -0,0 +1,211 @@
...
+## Appendix
+
+### Choose final validators
+
+The final validator set that take effect in tendermint is chosen at `end_block` event by:
+
+- Sort all the active validators by `voting_power desc, staking_address`
+- Take the first `max_validators` ones
+
+For example, if you are the `max_validators + 1`-th active validator, so you are not chosen yet, but if anyone in the
```
tomtau: it'll be helpful to show this example more in detail altogether with the expected generated ValidatorUpdates

docs/modules/staking-state.md
```diff
@@ -0,0 +1,211 @@
...
+struct StakedState {
+    address: StakedStateAddress,
+    nonce: u64,
+    bonded: Coin,
+    unbonded: Coin,
+    unbonded_from: Timespec,
+    validator: Option<Validator>,
+}
+
+struct Validator {
```
tomtau, line 13: the slashing docs say that, for byzantine faults, it'll be slashed just once -- how will that be decided / checked that it happened, so that it doesn't get slashed? My guess is that it'll be based on `jailed_until.is_some()`, but I may be wrong -- in any case, it'll be good to explicitly state this "interaction"

tomtau (changes requested):
looks good for a start

docs/modules/staking-state.md
```diff
@@ -0,0 +1,211 @@
...
+  - After a validator inactivated, the signing vote might still arrive for the next two blocks, we don't want to issue a
+    false warning in this case.
+  - Validator might quit and re-join very fast (by `UnbondTx`/`DepositTx`/`NodeJoinTx`), in this case it's liveness
+    tracking record is preserved.
+
+So the liveness tracker is only removed when validator record get [cleaned up](#clean-up).
+
+### Inactive validator re-join with different validator key
+
+When an inactive validator re-join, it can provide different validator key, but it still need to be hold responsible for
```
lezzokafka: Just want to check if I understand it correctly: 
- Validator A usign validator key K_1 and Staking address A_1 to join the network.
- Then he leave the network. 
- He rejoin with validator key K_2 and Staking address A_1 to rejoin andit will be slash if byzantine fault was committed under K_1  and A_1?

lezzokafka (commented):
Some comments:

docs/modules/staking-state.md
yihuang: Yes, if we agree on this version of logic.

docs/modules/staking-state.md
yihuang: There is a many to one `tm address -> staking address` map for all the validators, `old_validator_keys` is like the reverse of this map. So you can clean them up by reversed lookup like:
```rust
for key in validator.old_validator_keys {
  tm_address_map.remove(key.address())
}
```
Also, since no evidence the validator key is compromised, no need to blacklist it? 
And there is no cleanup logic for the blacklist map yet, so better to keep it as small as possible.
I think it's the most balanced solution.

docs/modules/staking-state.md
yihuang: I add a simple description under the "Implications of jailing":

"Jailed validators won't be slashed again for byzantine faults."

docs/modules/staking-state.md
yihuang: This is actually deviated from the current logic, there's no `EFFECTIVE_MIN_STAKE` in this one.

docs/modules/staking-state.md
yihuang: Yes, the evidence will include K_1, and we need to be able to connect it to A_1, and slash it.

docs/modules/staking-state.md
yihuang: There is a many-to-one `tm address -> staking address` map, the `old_validator_keys` is like the reverse of it, so when cleanup, we can iterate `old_validator_keys` to cleanup that map.

tomtau: Any bounds on that? As this gets loaded in memory, one can blow this up cheaply by repeated node joins with an inactive or a borderline active (do repeated small amount unbond/deposit)

docs/modules/staking-state.md
yihuang: Done

docs/modules/staking-state.md
tomtau: For the blacklist, I assume it's on-disk key-value lookup and it won't take much space with index key compressions and other magic the db may do. For faster lookups, one may have a fixed size probabilistic filter and only check on-disk to exclude false positives.

with this dual handling, there's an extra complexity and one needs to be careful. 
I'm fine with it -- but extra clarification needed: https://github.com/crypto-com/chain-docs/pull/102#discussion_r392129384

docs/modules/staking-state.md
yihuang: It's also trivial to switch to current logic by inactivating the validators who are not chosen into the top N at the end_block. But I think choose automatically seems a little bit better.

docs/modules/staking-state.md
yihuang: Actually, with this version, the validator needs to unbond to bellow `min_required_staking` to inactivate itself.
So there is some opportunity cost to it, but it brings us to another problem which is no cheap way to just change the validator key?

yihuang: > Any bounds on that?

yeah, maybe the simplest solution is set a bound on that, when reached the bound, he's not allowed to use a new key again, but still allow to change to one of the old keys he owns?

tomtau: there is an opportunity cost, but that assumes one joined in order to be a validator and cared about the long term -- what if I joined simply for DoS -- this dynamic vector is a good target.

tomtau: > yeah, maybe the simplest solution is set a bound on that, when reached the bound, he's not allowed to use a new key again, but still allow to change to one of the old keys he owns?

possibly? or banning old keys altogether + some time limit on update frequency?

later, it'll be good to add some "update" tx -- either to update just the non-consensus metadata (name, security contact) or rotate the consensus pubkey -- it doesn't need to be in the immediate scope of this PR, but it's good to keep in mind how roughly that could work / if something like this could be added without introducing some attack vectors

yihuang: > possibly? or banning old keys altogether

With banning, I suppose we still need some way to clean the old stuff? Maybe recording a timestamp for now? `tm address -> (staking address, ban_time)`.

> + some time limit on update frequency?

What do you mean by this?

yihuang: > later, it'll be good to add some "update" tx -- either to update just the non-consensus metadata (name, security contact) or rotate the consensus pubkey -- it doesn't need to be in the immediate scope of this PR, but it's good to keep in mind how roughly that could work / if something like this could be added without introducing some attack vectors

If we will support cheaply rotate the consensus pubkey, then the banned key will increase rapidly even for a outer storage. Then I think storing it with `StakedState` with a maximum bound would be simpler.
We can store the timestamp with it:
```
Vec<(tm address, timespec)>
```
So we can clean up old ones(> max_evidence_age) as soon as possible.

tomtau: >  time limit on update frequency

e.g. once per day and only above the min required stake can do it

docs/modules/staking-state.md
```diff
@@ -0,0 +1,254 @@
...
+  deliver_tx
+  - join_node(addr3, 3)
+  - join_node(addr4, 4)
+  active validators:
+  - addr1 -> 1
+  - addr2 -> 2
+  - addr3 -> 3
+  - addr4 -> 4
+  end_block validator updates:
+  - addr3 -> 3
```
tomtau: should here be also:
```
addr1 -> 0
addr4 -> 4
```

as 1 < 3 and 1 < 4?

(I don't know if the arrows in this example refer to vote power?)

tomtau (changes requested):
looks ok, but the example may have an error or needs a clarification

docs/modules/staking-state.md
yihuang: > > time limit on update frequency
> 
> e.g. once per day and only above the min required stake can do it

You mean frequency to update validator key? If we have upper bound on the length of the used key list (something like 100), then don't need a limit on frequency?

docs/modules/staking-state.md
yihuang: Yes, the right side of the arrow is the voting power, I should add some clarifications.
I made the mistake to chosen smaller voting power first here, I'll change.

yihuang: Done

docs/modules/staking-state.md
tomtau: size upper bound (could be even compile-time constant) + cleaning works as well.

PR #106, opened 2020-03-16
Problem (Fix #105): wrong command line argument to get another staking address

### LINKED ISSUE #105: Problem: wrong command line argument to get another staking address
https://crypto-com.github.io/getting-started/send_your_first_transaction.html#create-transfer-staking-address
it should be `--type Staking` instead of `--type Transfer` below the caption
```Run another time and you will get the address that stores the unbonded funds.```

allthatjazzleo:
Solution:
- fix command line argument


lezzokafka (approved):

lezzokafka (approved):
Nice catch Leo

PR #108, opened 2020-03-16
Problem: Banning validator pubkey is not necessary

yihuang:
Solution:
- Remove banning of double signed validator key.

### DIFF
--- a/docs/modules/staking-state.md
+++ b/docs/modules/staking-state.md
@@ -56,7 +56,7 @@ There are several variants of it:
 The only way to transit to active validator is by executing `NodeJoinTx`, the preconditions are:
 
 - `bonded >= min_required_staking`
-- The validator pubkey/address is not banned or already used
+- The validator pubkey/address is not already used by others, it's ok to re-use the old keys used by itself if it's a re-join from an inactive validator.
 - Not jailed if transitting from inactive validator
 
 ### From "active validator" to "inactive validator"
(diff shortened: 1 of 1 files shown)


docs/modules/staking-state.md
```diff
@@ -56,7 +56,7 @@ There are several variants of it:
 The only way to transit to active validator is by executing `NodeJoinTx`, the preconditions are:
 
 - `bonded >= min_required_staking`
-- The validator pubkey/address is not banned or already used
+- The validator pubkey/address is not already used
```
devashishdxt: If a validator manually unbonds and wishes to rejoin later, will they need a new pubkey/address?

yihuang: > If a validator manually unbonds and wishes to rejoin later, will they need a new pubkey/address?

It's ok to re-use your own keys, I just added a sentence to make it clear.

devashishdxt (commented):
Is there any reason for this change?

yihuang:
> Is there any reason for this change?

Because after discussing with @tomtau , we think do the banning in chain core don't help much, if our purpose is only to prevent operator making mistakes (like reusing a leaked key), we can do it in client.

tomtau:
> Is there any reason for this change?

making it simple -- some of blacklist use cases are not relevant; others are handled by that old key list: https://github.com/crypto-com/chain-docs/blob/master/docs/modules/staking-state.md#inactive-validator-re-join-with-different-validator-key

PR #110, opened 2020-03-19
Problem(Fix #76) : No explanation of reward configuration parameters

### LINKED ISSUE #76: Problem: no explanation of reward configuration parameters
Currently, it just says these are the parameters and this is how stuff is calculated: https://crypto-com.github.io/getting-started/staking.html#rewards
but it doesn't explain what setting each term does

lezzokafka:
**Solution**:

-`getting-started/reward-and-punishments.md` Added  the "Reward parameters explained" section. preview :
Screen Shot 2020-03-19 at 3 53 13 pm

**Extra**: 
Fixed some broken links and added links to` /modules/transactions.md`


tomtau (changes requested):
it's still not that clear -- I assume the example shown assumes fixed total stake?
and lacks many details -- e.g. r0 capping all years; difference between initial tau (tau_0?) and decayed tau... and the overall effect -- setting decay higher/lower, what happens to emitted rewards when total stake lower/higher

lezzokafka:
>difference between initial tau (tau_0?) and decayed tau... 

@yihuang 
I see one thing that would cause some confusion is the way we name this initial tau  (`monetary_expansion_tau`) maybe we should call it `monetary_expansion_tau0`, so that people won't mix it up with the actual value of `tau`.

yihuang:
> > difference between initial tau (tau_0?) and decayed tau...
> 
> @yihuang
> I see one thing that would cause some confusion is the way we name this initial tau (`monetary_expansion_tau`) maybe we should call it `monetary_expansion_tau0`, so that people won't mix it up with the actual value of `tau`.

The network parameter `monetary_expansion_tau` is just the initial value of tau, after initialization, just call it `tau`?

PR #116, opened 2020-03-27
Problem: nonce logic makes client difficult to handle

yihuang:
Solution:
- Make nonce to be the count of transactions sent from the staking address.

### DIFF
--- a/docs/modules/staking-state.md
+++ b/docs/modules/staking-state.md
@@ -180,26 +180,12 @@ Jailed validators won't be slashed again for byzantine faults detected in jailin
 
 ### Nonce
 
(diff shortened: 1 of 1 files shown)


docs/modules/transactions.md
```diff
@@ -82,6 +82,7 @@ struct ChainInfo {
   7. Check staking state for withdraw:
 
      - `nonce` match
+     - `amount` match (optional)
```
tomtau: with the simpler fee scheme, this shouldn't be a problem as the tx fails if unbonded != output sum + (computed) fee... instead of passing and paying some larger/smaller fee?

yihuang: Yes, it's effectively just output_sum.

PR #124, opened 2020-04-08
Problem: No docs for the new jellyfish merkle tree

yihuang:
Solution:
- Add one

### DIFF
--- a/docs/modules/jellyfish-merkle.md
+++ b/docs/modules/jellyfish-merkle.md
@@ -1,0 +1,296 @@
+# Jellyfish Merkle Trie
+
+> The ascii arts is copied from libra code comments.
+
+## Concept
+
+```rust
+type H256 = [u8; 32];
+fn hash(value: Option<Value>) -> H256;
+fn hash(left: H256, right: H256) -> H256;
+```
+
+We build the conceptual model first, then apply the optimizations on top of it to make it practical.
+
+- It's a complete binary tree with 256 depth, the leaf node contains `Option<Value>`, `None` means value not exists.
+- It's a binary trie which implements the map from  `H256` to `Option<Value>`.
+- It's a merkle tree, so all the internal nodes contains hashes of it's direct children.
+  - Hash of leaf nodes: `hash(value)`.
+  - `hash(None)` is a constant, it's called a placeholder hash.
+  - Hash of internal nodes: `hash(left, right)`.
+
+Every modification in leaf nodes should cause the root hash and all the internal nodes along the path to change.
+
+> `hash(left, right) != hash(right, left)`
+>
+> This is important to fulfill above property.
+
+All the nodes are stored in underlying key-value storage(usually on disk), and
+all the nodes committed into underlying storage are considered immutable, so we
+can query the old state, unless we decide to prune the states that is too old.
+
+### Merkle Proof
+
+- Membership merkle proof for a key is the list of sibling hashes along the
+  path from leaf node to the root node, combined with the key itself.
+- Non-membership proof is just an membership proof lead to a `None` value.
+- Range proof, TODO.
+
+## Optimization
+
+- Flatten a 4 layer complete binary tree into array of 16 leaf nodes, this reduces the total numbers of nodes.
+
+  The internal node only stores leaf nodes of this sub-tree, these leaf nodes might denote either a leaf node or an
+internal node in outer tree.
+
+  The hash of the internal node is the root hash of this subtree, which is computed like this:
+
+  ```text
+  InternalNode[0, 1, 2, 3, 4, 5, 6, 7, 8, 9, A, B, C, D, E, F]
+
+    4 ->              +------ root hash ------+
+                      |                       |
+    3 ->        +---- # ----+           +---- # ----+
+                |           |           |           |
+    2 ->        #           #           #           #
+              /   \       /   \       /   \       /   \
+    1 ->     #     #     #     #     #     #     #     #
+            / \   / \   / \   / \   / \   / \   / \   / \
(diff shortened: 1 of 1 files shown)


docs/modules/jellyfish-merkle.md
```diff
@@ -0,0 +1,250 @@
...
+type H256 = [u8; 32];
+fn hash(value: Option<Value>) -> H256;
+fn hash(left: H256, right: H256) -> H256;
+```
+
+We build the conceptual model first, then apply the optimizations on top of it to make it practical.
+
+- It's a complete binary tree with 256 depth, the leaf node contains `Option<Value>`, `None` means value not exists.
+- It's a binary trie which implements the map from  `H256` to `Option<Value>`.
+- It's a merkle tree, so all the internal nodes contains hashes of it's direct children.
```
tomtau, line 17:
```suggestion
- It's a merkle tree, so all the internal nodes contains hashes of its direct children.
```

docs/modules/jellyfish-merkle.md
```diff
@@ -0,0 +1,250 @@
...
+>
+> This is important to fulfill above property.
+
+All the nodes are stored in underlying key-value storage(usually on disk), and all the nodes committed into underlying
+storage are considered immutable, so we can query the old state, unless we decide to prune the states that is too old.
+
+### Merkle Proof
+
+- Inclusion merkle proof is the list of sibling hashes along the path from leaf node to the root node.
+- Exclusion proof is just an inclusion proof of value `None`.
```
tomtau: but one should also verify the "position" of `None`? I assume any leaf is potentially missing, so there may be multiple possible "None" values?

docs/modules/jellyfish-merkle.md
```diff
@@ -0,0 +1,250 @@
...
+                |           |           |           |
+    2 ->        #           #           #           #
+              /   \       /   \       /   \       /   \
+    1 ->     #     #     #     #     #     #     #     #
+            / \   / \   / \   / \   / \   / \   / \   / \
+    0 ->   0   1 2   3 4   5 6   7 8   9 A   B C   D E   F
+    ^
+  height
+  Note: # denotes the hash of its two direct children
+  ```
```
tomtau: I guess it'll be helpful to show the flattened tree?
Is it:
```
root: InternalNode[0,1,2,3,4,5,6,7,8,9,A,B,C,D,E,F]
```
?

docs/modules/jellyfish-merkle.md
```diff
@@ -0,0 +1,250 @@
...
+    3 ->        +---- # ----+           +---- # ----+
+                |           |           |           |
+    2 ->        #           @           8           #
+              /   \                               /   \
+    1 ->     0     3                             #     @
+                                                / \
+    0 ->                                       C   @
+    ^
+  height
+  Note: @ denotes placeholder hash.
```
tomtau: This is a bit puzzling, as the above thing doesn't say it's a sub-tree -- one can also interpret that as a complete tree with 16 leafs

 I guess that's taken from: https://github.com/libra/libra/blob/master/storage/jellyfish-merkle/src/node_type/mod.rs#L198

So, it'd require more explanation -- I guess here the assumption is: "assume that the above tree is showing a subtree 0, 3, 8 are leafs and C is an internal node, everything else is empty"?

If C is the internal node, does it mean it's something like:
```
InternalNode[0,@,1,@,4,5,6,7,8,9,A,B,C,D,E,F]
```

docs/modules/jellyfish-merkle.md
tomtau: Or this can be reformulated as these are proofs for "Key" -- so the proof for that key either leads to a value or None

docs/modules/jellyfish-merkle.md
```diff
@@ -0,0 +1,250 @@
...
+### `get`
+
+```rust
+// proof generation is ignored.
+fn get(key, version) -> Option<Value> {
+  let mut next_key = NodeKey::new(version, vec![]);
+  for i in 0..64 {
+    match get_node(next_key) {
+      Internal(node) => {
+        let nibble = key.nibbles(i);
```
tomtau: for these simplified examples, it'll be good to explain what `.nibbles` does

tomtau (changes requested):
looks good, but a bit terse at times -- I had to go to the original Libra comments to understand some of the examples.
this https://eprint.iacr.org/2018/955.pdf also looks a bit similar approach?

docs/modules/jellyfish-merkle.md
yihuang: It's actually:
```
InternalNode {
  0 -> child(is_leaf=true),
  3 -> child(is_leaf=true),
  8 -> child(is_leaf=true),
  C -> child(is_leaf=false)
}

docs/modules/jellyfish-merkle.md
yihuang: > but one should also verify the "position" of `None`? I assume any leaf is potentially missing, so there may be multiple possible "None" values?

Yes, the list of siblings need to be combined with the key itself, I'll update.

yihuang:
> this https://eprint.iacr.org/2018/955.pdf also looks a bit similar approach?

I only taken a glance at that, it seems more like a binary search tree without re-balance?

docs/modules/jellyfish-merkle.md
yihuang: Yes, I added that.

docs/modules/jellyfish-merkle.md
yihuang: Done

tomtau (approved):
looks good for start

PR #137, opened 2020-05-13
Problem: no test vectors for different transaction types (fixes #68)

### LINKED ISSUE #68: Problem: no test vectors for different transaction types
Just like in BIPs:
https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#Test_Vectors
https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki#Test_vectors
https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki#Test_Vectors
https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki#Test_Vectors_and_Reference_Code

it'll be good to have some test vectors for transaction data payload + witness + addresses.
There could be a command in dev-utils "test-vectors" that will generate these fixed test vectors.
(It could also later be added to CI to check transaction format wasn't broken.)

tomtau:
Solution: generated a few test vectors with modified dev-utils and
documented them

### DIFF
--- a/docs/getting-started/serialization.md
+++ b/docs/getting-started/serialization.md
@@ -3,3 +3,155 @@
 After several iterations of binary formats, Crypto.com Chain settled on using the SCALE (Simple Concatenated Aggregate Little-Endian) codec.
 
 It is formally described in Section B.1 of [Polkadot RE Protocol Specification](https://github.com/w3f/polkadot-re-spec/blob/master/runtime-environment-spec/polkadot_re_spec.pdf).
+
+## Test Vectors
+### Key pairs and Addresses
+Transaction-related keys use the secp256k1 ECC, as in Bitcoin, Ethereum etc.
+The following test vectors were generated with the seed `ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff`.
+
+#### View key pair
+A secp256k1 key pair used for authentication when requesting transaction data from enclaves.
+
+Secret: `b4518a5a1594816446f9a50390fe2d5a896b91590b794bcd7ff4bee75d2ab601`
+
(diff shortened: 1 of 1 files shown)


yihuang (commented):
How these test vector should be used?

tomtau:
> How these test vector should be used?

when one creates an independent implementation (e.g. on embedded devices where one doesn't / can't use the heavy client-* stuff with a myriad of dependencies), one can refer to this for basic cross-checking that given the same inputs (keypairs + transaction arguments), the implementation generates the same outputs... it could be included in automated compliance tests etc.
One would still want to test on the public testnet that the created transactions are broadcasted and valid, so these test vectors are more for development.

docs/getting-started/serialization.md
```diff
@@ -3,3 +3,154 @@
...
+    
+##### Deposit
+Here, the witness payload will be obfuscated, the transaction payload is public.
+Example for plain payload:
+* spending output 0 of transaction `d0a35414d432dc757433e01a1523a53d34ef94dfaee09fd62564f42c783a5ea1`: `d0a35414d432dc757433e01a1523a53d34ef94dfaee09fd62564f42c783a5ea10000`
+* witness and attribute payloads shown above
+* plain transaction payload: `04d0a35414d432dc757433e01a1523a53d34ef94dfaee09fd62564f42c783a5ea1000000410afb0ccf51aefd111bceafb6c298acd4decaf6002a0100000000000000` (txid: `510f78adeae403e56b40f5c3bea1ff70f3dec2da6911e1e9b1f0ea03203019bc`)
```
yihuang, line 141: One still need the staking address to generate the tx payload?

tomtau: a good catch -- added it

yihuang (approved):

leejw51crypto (approved):
lgtm

Issue #141, opened 2020-05-19
Problem: MLS member leaving/joining interaction with council nodes not documented


tomtau:
From the audit discussion
> Who among the validators invites a newcomer when a new validator node joins?
Assuming TDBE is (in normal circumstances) provided with network updates (relevant transactions, events...), it'll go like this:
block N (time N_t), there's a valid NodeJoinTx with a metadata payload, including a ClientInitKey
TDBE corresponding to node with index 0 in the ratchet tree generates "Welcome" and "Send" payloads wrapped in a single MLS-related transaction that's then broadcasted/gossiped by Tendermint (let's call it MlsAddTx)
block M (time M_t), MlsAddTx is proposed to be in a block:
if M_t <= N_t + timeout, it's considered valid and propagated to validator TDBEs
if M_t > N_t + timeout, it's considered invalid and the process restarts with Step 2., but TDBE corresponding to a node with index 1
(and this may continue with indices 2,3,... the  consensus assumption is <1/3 of nodes are malicious)
One consequence of this is there may be multiple MLS Welcome/Send payloads for one newcomer -- only one of them is applied.

> Who among the validators triggers the group and key update when a validator node leaves?
Same logic as above with joining, but there multiple ways to trigger "validator node leaving" (operator decides to leave and broadcasts relevant transaction, its node was observed to be faulty, ...)

I guess for the nodes that are not in the validator set, but requested to join / have enough stake, their TDBE keypackage will still be a part of the MLS group structure and expected to generate those MlsAddTx transactions/react to the network changes.

For nodes not generating these update Txs in time, they may need to be punished

tomtau:
Also one thing TBD -- how's it done during genesis. I guess options are:
1) one of the initial validators is expected to generate the MLS payloads prior to genesis.json creation and they'll be dumped into genesis.json and processed by initial validators
2) they'll be just keypackages, as now, and one of the initial validators will be expected to generate and broadcast a TX with MLS payloads post-network launch within some time limit (following the same/similar logic as with joining/leaving)

yihuang:
> Also one thing TBD -- how's it done during genesis. I guess options are:
> 
> 1. one of the initial validators is expected to generate the MLS payloads prior to genesis.json creation and they'll be dumped into genesis.json and processed by initial validators
[...]

PR #152, opened 2020-06-12
Problem: MLS member leaving/joining interaction with council nodes not documented (WIP #141 fixes #142)

### LINKED ISSUE #142: Problem: unnecessary complexity for obfuscation key forward secrecy
Currently, the protocol in the implementation plan: https://github.com/crypto-com/chain-docs/blob/master/plan.md#key-generation-and-rotation
describes 2-steps that all nodes will be expected to generate random payloads, broadcasts signed commitments to those payloads, and then broadcast those payloads.

This may be simplified a bit, some ideas:

- "0 steps" (but with no auth outside the group -- but it may be fine): latest MLS draft defines an "exporter": https://tools.ietf.org/html/draft-ietf-mls-protocol-09#section-7.11 -- similar to https://tools.ietf.org/html/rfc5705 so it can be used for a keying material -- e.g. new obfuscation key could be just defined as something like `MLS-Exporter("tx_obfuscation_key", block_height_this_was_initiated, AES_GCM_SIV_KEY_LEN)` ... so gets refreshed every time the group changes/updates

- "1 step": https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-vrf-06#section-5.1 -- generate a key pair using "MLS-Exporter"; public key broadcasted outside; ECVRF_prove to generate a secret... or each node having a dedicated key pair and then sending outputs of ECVRF_prove as application messages.

...
Will need to think about it more, but 0 step may be fine

tomtau:
Solution: sketched out a doc containing more details beyond the original implementation plan
(it still lacks some parts, so still WIP)
+ some parts -- e.g. new transaction types -- can be moved to other modules when it's agreed on and detailed/polished later
--
also, as the MLS protocol draft changed since the original implementantion plan,
key rotation procedure was simplified to make use of MLS exporter functionality

### DIFF
--- a/docs/modules/tdbe.md
+++ b/docs/modules/tdbe.md
@@ -1,0 +1,247 @@
+# Transaction Data Bootstrapping Enclave details
+
+Transaction Validation Enclave (TVE):
+- responsible for validating Enclave tx types;
+- holds current tx obfuscation key;
+- reacts to *immediate* requests (chain-abci, i.e. potentially uncommitted or invalid transaction; encryption requests from tx-query enclave)
+- chain-abci directly talks to it during (consensus) state-machine execution
+
+Transaction Data Bootstrapping Enclave (TDBE):
+- responsible for generating MLSHandshake tx types;
+- fetches old transaction data payloads;
+- prerequisite for a node administrator to submit a node join tx request or construct genesis.json
+- reacts to *block-committed* requests:
+  - needs to run a Tendermint light client verification (TODO: trusted anchor injected as compile-time data?) and be connected to its Tendermint node RPC
+  - chain-abci does not talk to it (it may invoke it during startup / for fetching old tx data, but can't talk to it during state-machine execution)
+  - "knows" when it should generate MLSHandshake tx: 1) when in mid-time of keypackage expiration; 2) being "leftmost" node 
+  - based on valid block-committed requests, it should update its internal states, derive secrets and push tx obfuscation key to TVE
+
+
+## new/rejoining node
+(for non-genesis specified ones)
+- submit council or community node join request TX
+- if valid, wait until CommitChangeTx is committed with Welcome payload
+- mark the block where this CommitChangeTx is committed as the "fetched up to" block
+and obtain old transaction data
+
+TODO: fetch keypackages/leaf nodes? https://github.com/mlswg/mls-protocol/issues/344
+
+### Obtaining old transaction data
+When a node is started up (before tendermint consensus syncing etc.), it should be provided
+"genesis-skip" (in the case it's one of nodes starting at genesis)
+or one or more connection strings
+that would proxy to some public full node RPC.
+
+Via the Tendermint light-client-verifying connection, it should obtain the list of 
+needed transaction IDs (of withdrawal and transfer transactions):
+
+* full node / historical querying: all IDs between its last processed block and the last block on the remote node
+* validator: new IDs in the "UTXO set diff" between its last processed block and the last block on the remote node
+(TODO: https://github.com/crypto-com/chain/issues/794 ; initial implementation here can start with just "all IDs" for both)
+
+From remote node RPC, it should also learn its TDBE connection details
+and then initiate enclave-to-enclave mutually attested TLS: TODO https://github.com/crypto-com/chain/issues/1549
+and fetch the needed transaction payloads and seal them for its host.
+
+TODO: handling in abci -- options (probably start with option 1):
+1. record the "fetched up to" block;
+report to TM still the last processed block
+and run normal TM block-by-block syncing and skip tx-validation-enclave until the "fetched up to" block
+2. TM 0.34+ -- run state-sync for jellyfish merkle / staking state https://docs.tendermint.com/master/spec/abci/apps.html#state-sync
+and resume from the "fetched up to" block?
+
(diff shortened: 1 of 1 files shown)


docs/modules/tdbe.md
```diff
@@ -0,0 +1,221 @@
...
+}
+```
+
+##### TDBE handling / generation
+TDBE runs as a standalone enclave which includes TM light client and reacts to information received from it.
+It opens mutually attested TLS connection to tx-validation enclave (TVE) for pushing exported obfuscation keys.
+
+###### Commits
+When a block is committed with nodejoin, unbond or punishment events (liveness/byzantine issues, bue also including expired keypackages), 
+Node's TDBE corresponding to the leftmost non-empty leaf should generate and broadcast `CommitChangeTx` with proposals reflecting the triggering block's change (Add/Remove).
```
yihuang, line 180: when this leftmost non-empty leaf node non-live, it's very harmful, maybe using a special official node to do this job?
or more generally, how to detect the non-functioning of this node, and punish it?

tomtau: https://github.com/crypto-com/chain-docs/pull/152/files#diff-3c337734c61cba88aa44ecd090a7ac6dR215

tomtau: https://github.com/crypto-com/chain-docs/pull/152/files#diff-3c337734c61cba88aa44ecd090a7ac6dR169

tomtau: >  how to detect the non-functioning of this node

one knows when the triggering block arrived, so knows within what time to receive `CommitChangeTx`?

docs/modules/tdbe.md
```diff
@@ -0,0 +1,221 @@
...
+
+If CommitChangeTx is not received in a block with time less the triggering block's time + timeout,
+OR received CommitChangeTx was invalid;
+
+node's TDBE corresponding to the next leftmost non-empty leaf should generate and broadcast `CommitChangeTx` -- it should include
+a Remove proposal for the previous node(s) (that failed to submit in time or submitted invalid `CommitChangeTx`;
+it should be indicated in timeout-committed block
+);
+and include any additional proposals (if any) since the original triggering block;
+(it should take the latest "timeout" block's number as AAD.)
```
yihuang, line 195: timeout: 5 blocks

- block `n` triggers, v1 not respond to it
- block `n + 3` triggers, v1 gen `CommitChangeTx` for it, I guess this tx should be considered invalid if it not including both block `n` and `n+3`?
- block `n + 5` triggers, also block `n` timeout detected, v2 take over, it gen `CommitChangeTx` for blocks `n`, `n+3`, `n+5`, and remove v1.

it work like this right?

tomtau: timeouts may be in timespec... but perhaps block numbers may be easier to reason about for / to track what changes need to be included? (assuming the timeout is sufficiently large to prevent accidental timing out due to gossip delay / accidentally many new sudden blocks)

I'd expect under normal circumstances / operation, `v1` should observe all changes and respond immediately e.g.:
* block `n` triggers -> v1 gen `CommitChangeTx`
* block `n + 3` triggers ... -> v1 takes note of it
* block `n + 4` includes `CommitChangeTx` -> v1 gen `CommitChangeTx` for `n+3`
...

For that scenario you described:
>  I guess this tx should be considered invalid if it not including both block n and n+3?

Yes, the expectation would be to observe `n` `CommitChangeTx` first.

> it work like this right?
 
Yes, the v2 takeover is expected like that.

docs/modules/tdbe.md
```diff
@@ -0,0 +1,236 @@
+# Transaction Data Bootstrapping Enclave details
+
+Transaction Validation Enclave (TVE):
+- responsible for validating Enclave tx types;
+- holds current tx obfuscation key;
+- reacts to *immediate* requests (chain-abci, i.e. potentially uncommitted or invalid transaction; encryption requests from tx-query enclave)
```
devashishdxt, line 6:
```suggestion
- reacts to *immediate* requests (chain-abci, i.e. potentially uncommitted or invalid transaction; encryption and decryption requests from tx-query enclave)
```

leejw51crypto (approved):
lgtm

docs/modules/tdbe.md
yihuang: - block n triggers -> v1 gen CommitChangeTx tx1 and broadcast
- block n+1 triggers -> v1 gen CommitChangeTx tx2 and broadcast

tx1 and tx2 will go through tendermint consensus to get into blocks, I think it's possible that tx2 included in block before tx1? abci need to be able to keep tx2 for a while, after it also sees tx1, then process tx1 and tx2 in order?

devashishdxt (approved):

docs/modules/tdbe.md
tomtau: decryption requests don't need to be handled by TVE if it remains like now, where TVE and TQE have the same ProductId etc., so the sealed data can be unsealed by both... so it TQE can just be passed the sealed payloads from the runner/chain-abci

docs/modules/tdbe.md
tomtau: v1 needs to delay generating and broadcasting tx2 until it sees tx1 to be committed/included in a block (in time)

PR #153, opened 2020-06-12
Workaround for a broken link

juno-yu:
I don't find a high resolution version of original file online but this might make it barely readable


lezzokafka:
Thanks very much @juno-yu! 
I think the original post was from https://blog.cosmos.network/tendermint-in-a-nutshell-39d9f7f66ad7 ( and the ref link is already dead, as you have spotted. ) 
Maybe the [diagram](https://docs.tendermint.com/master/assets/img/tm-transaction-flow.258ca020.png) in this [tendermint docs](https://docs.tendermint.com/master/introduction/what-is-tendermint.html#stake) will be helpful? 

Thanks for your contribution, once again :D !

juno-yu:
nice this is of high definition, updated to image link

lezzokafka (approved):
Wonderful, thanks again @juno-yu !

PR #155, opened 2020-06-19
Problem: node join test vector doesn't contain a keypackage (fixes #146)

### LINKED ISSUE #146: Problem: node join test vector doesn't contain a keypackage
+ description of the expected structure referring to the IETF doc.

https://github.com/crypto-com/chain/pull/1684/files#diff-f82b1fb372c07b9409f261820dccebc6R325

tomtau:
Solution: updated the node join tx with the payload generated by dev-utils
with a sample keypackage


leejw51crypto (approved):
lgtm

PR #162, opened 2020-06-30
Problem: (Fix #161) Local full node deployment pages are not up-to-date and have typos

### LINKED ISSUE #161: Problem: Local full node deployment is not up-to-date
"Local full node deployment" and "Send Your First Transaction" is not up-to-date and has some typos.

calvinlauyh:
Solution: Update the local full node deployment and send first transaction page

Blocked by: https://github.com/crypto-com/chain/pull/1850

----
Special credits to @diereysaa who spotted the typos and gave us feedback. Thank you!


docs/getting-started/README.md
```diff
@@ -134,17 +140,24 @@ $ source ~/.profile
 Prepare initial chain data and try to install Intel SGX if the SGX device is not ready.
 
 ```
-$ make prepare
+$ sudo make prepare
```
tomtau: why sudo? is it running some permission commands? wouldn't it be better if sudo was used inside the makefile where needed rather than the whole make running under sudo?

docs/getting-started/send_your_first_transaction.md
```diff
@@ -87,13 +116,14 @@ Enter authentication token:       // Input the Authentication token
...
-// Leave blank because this tx is in same wallet
+# Leave blank because this tx is in same wallet
+Transaction successfully created
 ```
 
 - Then, you can `sync` and check `balance` of your wallet:
 
 ```bash
-$ ./target/debug/client-cli sync --name Default
+$ ./target/debug/client-cli sync --name Default --disable-fast-forward
```
tomtau, line 126: maybe fast forward shouldn't be a default if it causes issues + it's not secure (if one doesn't run a full node) / can't be used with the light client verification?

tomtau (commented):
looks ok, just a few comments -- not sure why `sudo` is needed? or is it because of the docker? if so, why not just adding the user to docker group?

docs/getting-started/README.md
calvinlauyh: The `sudo` is needed for copying the config files
```
copy devnet tendermint config
cp docker/config/devnet/tendermint/config.toml /tmp/data/tendermint/config/
cp: cannot create regular file '/tmp/data/tendermint/config/config.toml': Permission denied
Makefile:106: recipe for target 'init-tendermint' failed
make: *** [init-tendermint] Error 1
```

The default data file path is `/tmp/data`. but the `/tmp` folder is owned by the root.

docs/getting-started/send_your_first_transaction.md
calvinlauyh: The main reason driving me to change this is because the Devnet have zero fee and is affected by https://github.com/crypto-com/chain/issues/794

After the issue is fixed we could use the default of fast forward.

docs/getting-started/send_your_first_transaction.md
```diff
@@ -26,7 +36,9 @@ Confirm mnemonic:       // Confirm the mnemonic
 
 Authentication token: b74ce4590ebc9d3c2a3adace926304384ae9451f43560c9702402be53811da8b
 ```
-Then you will get the `Authentication token`, keep the token safe and it will be needed for all authorised commands.
+Then you will get the `Authentication token`, keep the token safe and it will be needed for all authorized commands.
+
+A pre-created Hierarchical Deterministic(HD) Wallet wallet mnemonic with funds inside are prepared for you in Devnet. You can find it [here](https://github.com/crypto-com/chain/blob/master/docker/config/devnet/tendermint/mnemonics.txt).
```
lezzokafka:
```suggestion
A pre-created Hierarchical Deterministic(HD) Wallet wallet mnemonic with funds inside are prepared for you in Devnet. Kindly restore the wallet by using this [mnemonic](https://github.com/crypto-com/chain/blob/master/docker/config/devnet/tendermint/mnemonics.txt) before moving on to the next step. 
```

lezzokafka (commented):
Looks great, just left a quick comment.
Thanks @calvinaco!

docs/getting-started/README.md
tomtau: what about just adding a step with sudo creating the folders + chown/chrgp?

calvinlauyh: Updated to separate commands:
```
sudo make create-path
sudo make chown-path user=... group=...
make prepare
```

lezzokafka (approved):

tomtau (changes requested):
looks ok, but should the PR be now first against `dev`? @allthatjazzleo

allthatjazzleo:
> looks ok, but should the PR be now first against `dev`? @allthatjazzleo

yes, should create the PR against `dev`

calvinlauyh:
> looks ok, but should the PR be now first against `dev`? @allthatjazzleo

Changed base branch to `dev`

tomtau (approved):

PR #179, opened 2020-07-16
Problem: MLS member leaving/joining interaction with council nodes not documented (WIP #141)

tomtau:
Solution:
- revised the proposed design, so that joining is immediate (node fetches MLS handshakes directly from remote TDBE instead of waiting for commit message)
- sketched out basic types for NACK
- sketched out a state diagram and explanation of transitions with regards to different events and MLS handshake messages

### DIFF
--- a/docs/modules/tdbe.md
+++ b/docs/modules/tdbe.md
@@ -19,10 +19,19 @@ Transaction Data Bootstrapping Enclave (TDBE):
 
 ## new/rejoining node
 (for non-genesis specified ones)
+- obtain old data (see below) from external node TDBE
+- request "invitation" from the external node TDBE:
+  - request should contain the merkle proof of the new node's associated staking state + keypackage
+  - responding node should verify both -- if correct, it'll generate `Welcome`, `Add` proposal, and `Commit`
+  - requesting node should verify the response -- if incorrect, start again (e.g. with a different external node TDBE on the list)
+- construct council or community node join request TX payload -- get it signed by the node operator
 - submit council or community node join request TX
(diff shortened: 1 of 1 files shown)


docs/modules/tdbe.md
```diff
@@ -19,10 +19,19 @@ Transaction Data Bootstrapping Enclave (TDBE):
 
 ## new/rejoining node
 (for non-genesis specified ones)
+- obtain old data (see below) from external node TDBE
+- request "invitation" from the external node TDBE:
```
yihuang, line 23: it should request "invitation" from the left-most leaf node, right?

tomtau: No, any node should do here, as add is triggered immediately by the newly joining node / operator (unlike remove which is triggered by commit of a block with some event / transaction)

tomtau: it makes sense to request it from the external one it was fetching data from

tomtau: any concerns if it is any node?

yihuang (approved):

docs/modules/tdbe.md
```diff
@@ -19,10 +19,19 @@ Transaction Data Bootstrapping Enclave (TDBE):
...
+  - request should contain the merkle proof of the new node's associated staking stake + keypackage
```
devashishdxt: What is `staking stake`?

devashishdxt (approved):

docs/modules/tdbe.md
tomtau: a typo, it should be "staking state"

PR #204, opened 2020-08-03
Problem: New TDBE related genesis parameters are not described in detail

### LINKED ISSUE #194: Problem: new genesis parameters (TDBE-related) are not described in detail

devashishdxt:
Solution: Added proper names and descriptions of TDBE related genesis parameters. Fixes #194.

### DIFF
--- a/docs/modules/tdbe.md
+++ b/docs/modules/tdbe.md
@@ -123,13 +123,16 @@ pub struct CommunityNode {
 ```
 
 ### extra network params
-- community node minimal required stake
-- mls handshake commit timeout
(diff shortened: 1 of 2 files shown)


tomtau (changes requested):
there is also https://chain.crypto.com/docs/protocol/network-parameters.html where they should be reflected

devashishdxt:
> there is also https://chain.crypto.com/docs/protocol/network-parameters.html where they should be reflected

Done

tomtau (approved):

PR #207, opened 2020-08-04
Problem: Details of catch-up process and data bootstraping are missing

devashishdxt:
Solution: Updated the "Catch-up process" section in docs. Related to #195.

Note: The other half of #195, i.e., invitation protocol will be done in a separate PR.

### DIFF
--- a/docs/modules/tdbe.md
+++ b/docs/modules/tdbe.md
@@ -35,29 +35,68 @@ and obtain old transaction data (if any)
 
 TODO: fetch keypackages/leaf nodes? https://github.com/mlswg/mls-protocol/issues/344
 
-### Obtaining old transaction data
-When a node is started up (before tendermint consensus syncing etc.), it should be provided
-"genesis-skip" (in the case it's one of nodes starting at genesis)
-or one or more connection strings
-that would proxy to some public full node RPC.
+### Catch-up process
 
-Via the Tendermint light-client-verifying connection, it should obtain the list of 
-needed transaction IDs (of withdrawal and transfer transactions):
+When a node starts, it should first check the configuration parameter `perform_catchup`. If `perform_catchup` is `true`,
+the node should perform following catch-up process to fetch all the changes in UTXO set from `last_block_height` of
+curent node to `current_block_height` of remote node. If `perform_catchup` is `false`, it should skip the catch-up
+process and start performing normal operations.
 
-- full node / historical querying: all IDs between its last processed block and the last block on the remote node
-- validator: new IDs in the "UTXO set diff" between its last processed block and the last block on the remote node
-(TODO: https://github.com/crypto-com/chain/issues/794 ; initial implementation here can start with just "all IDs" for both)
+To obtain all the changes in UTXO set from `last_block_height` of current node to `current_block_height` on remote node,
+it can first use Tendemint light client verifying connection to obtain a list of `Withdraw` and `Transfer`
+`transaction_id`s between the two block heights. Once all the transaction IDs are available, it can start establishing
+an enclave-to-enclave mutually attested TLS stream with remote TDBE server to get all the `TxWithOutputs` corresponding
+to above `transaction_id`s. After fetching all the transactions from remote TDBE server it can directly persist them
+using `chain-storage` (using [`chain_storage::store_sealed_log`](https://github.com/crypto-com/chain/blob/e24f29fceabebb979ac4ba1947148040d0be1cc5/chain-storage/src/api.rs#L152);
+TODO: Should this process also include creating and spending UTXOs or should that be done during tendermint
+synchronization process?) and update `last_fetched_block` in `chain-abci`.
 
-From remote node RPC, it should also learn its TDBE connection details
-and then initiate enclave-to-enclave mutually attested TLS: TODO https://github.com/crypto-com/chain/issues/1549
-and fetch the needed transaction payloads and seal them for its host.
+After all the transaction data is persisted, `chain-abci` can start listening to ABCI requests block-by-block and skip
+`tx-validation` for `Transfer` and `Withdraw` transactions until `last_fetched_block`. `chain-abci` should still do
+following basic validations for each transaction in a block:
 
-TODO: handling in abci -- options (probably start with option 1):
-1. record the "fetched up to" block;
-report to TM still the last processed block
-and run normal TM block-by-block syncing and skip tx-validation-enclave until the "fetched up to" block
-2. TM 0.34+ -- run state-sync for jellyfish merkle / staking state https://docs.tendermint.com/master/spec/abci/apps.html#state-sync
-and resume from the "fetched up to" block?
+- If `block_height <= last_fetched_block` **AND** (`tx_type == Transfer` **OR** `tx_type == Withdraw`) **AND**
+  transaction is present in storage, then, check if outputs size matches and all the inputs are unspent.
+- If `block_height <= last_fetched_block` **AND** (`tx_type == Transfer` **OR** `tx_type == Withdraw`) **AND**
+  transaction is not present in storage, then, skip the transaction as it is an invalid transaction.
+- If `block_height <= last_fetched_block` **AND** (`tx_type != Transfer` **AND** `tx_type != Withdraw`), then, perform
+  normal transaction validation.
+- If `block_height > last_fetched_block`, then, perform normal transaction validation.
+
+:::tip Note:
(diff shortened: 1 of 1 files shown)


docs/modules/tdbe.md
```diff
@@ -35,29 +35,45 @@ and obtain old transaction data (if any)
...
+listen for ABCI requests from tendermint. _Genesis skip_ includes changes in UTXO set from `last_block_height` of
+current node to `current_block_height` on remote node.
+
+To obtain all the changes in UTXO set from `last_block_height` of current node to `current_block_height` on remote node,
+it can first use Tendemint light client verifying connection to obtain a list of `Withdraw` and `Transfer`
+`transaction_id`s between the two block heights. Once all the transaction IDs are available, it can start establishing
+an enclave-to-enclave mutually attested TLS stream with remote TDBE server to get all the `TxWithOutputs` corresponding
+to above `transaction_id`s. After fetching all the transactions from remote TDBE server it can directly persist them
+using `chain-storage` (using [`chain_storage::store_sealed_log`](https://github.com/crypto-com/chain/blob/e24f29fceabebb979ac4ba1947148040d0be1cc5/chain-storage/src/api.rs#L152);
+TODO: Should this process also include creating and spending UTXOs or should that be done during tendermint
```
tomtau, line 51: > TODO: Should this process also include creating and spending UTXOs or should that be done during tendermint synchronization process?

Since here, it'll follow block-by-block, it'll still need to know how to reply when -- so probably needs to temporarily mark for these prefetched transactions the expected block height when they are created?

docs/modules/tdbe.md
```diff
@@ -35,29 +35,45 @@ and obtain old transaction data (if any)
...
+synchronization process?) and update `last_fetched_block` in `chain-abci`.
+
+After all the transaction data is persisted, `chain-abci` can start listening to ABCI requests block-by-block and skip
+`tx-validation` until `last_fetched_block`. `chain-abci` should still do a basic validation and update UTXOs.
```
tomtau: only tx-validation of withdraw/transfer

docs/modules/tdbe.md
```diff
@@ -35,29 +35,45 @@ and obtain old transaction data (if any)
...
+
+To obtain all the changes in UTXO set from `last_block_height` of current node to `current_block_height` on remote node
+using tendermint light client, we use strategy similar to the one used in `client-core` for synchronizing the client.
+
+- Using `LightClient::verify_to_highest()`, retrieve `current_block_height`, `app_hash` and `block_hash` (of last block)
+  of remote node.
+- Split the block heights into a list of smaller chunks (`chunk_size` can be configurable), starting from
+  `last_block_height` of current node uptil `current_block_height` received from light client, and for each `chunk` of
+  `block_height`s:
+  - Call `block_results` on remote node's tendermint RPC for each chunk. For each `block_results` in the response:
```
tomtau: verify  `transaction_id`s / block_results against app hash?

docs/modules/tdbe.md
```diff
@@ -35,29 +35,45 @@ and obtain old transaction data (if any)
...
-1. record the "fetched up to" block;
-report to TM still the last processed block
-and run normal TM block-by-block syncing and skip tx-validation-enclave until the "fetched up to" block
-2. TM 0.34+ -- run state-sync for jellyfish merkle / staking state https://docs.tendermint.com/master/spec/abci/apps.html#state-sync
-and resume from the "fetched up to" block?
+### Catch-up process
+
+When a node is started up, it should be provided with a _genesis skip_ to catch up with other nodes before starting to
+listen for ABCI requests from tendermint. _Genesis skip_ includes changes in UTXO set from `last_block_height` of
+current node to `current_block_height` on remote node.
```
tomtau: the original meaning in that sentence was "genesis skip" was as in some flag to denote "skip this / don't do this -- this node is being started up at genesis, no need for any catch up", so this revised sentence is the opposite

tomtau (changes requested):

docs/modules/tdbe.md
devashishdxt: Changed.

docs/modules/tdbe.md
devashishdxt: Added verification step.

docs/modules/tdbe.md
devashishdxt: Changed.

docs/modules/tdbe.md
devashishdxt: Can you elaborate? I'm not sure if I understand what you mean correctly.

docs/modules/tdbe.md
tomtau: shouldn't that step be before fetching them / putting them in a chunk? if txid verification with `app_hash` doesn't match, why would want even want to continue fetching from that node?

docs/modules/tdbe.md
```diff
@@ -35,29 +35,49 @@ and obtain old transaction data (if any)
...
+`transaction_id`s between the two block heights. Once all the transaction IDs are available, it can start establishing
+an enclave-to-enclave mutually attested TLS stream with remote TDBE server to get all the `TxWithOutputs` corresponding
+to above `transaction_id`s. After fetching all the transactions from remote TDBE server it can directly persist them
+using `chain-storage` (using [`chain_storage::store_sealed_log`](https://github.com/crypto-com/chain/blob/e24f29fceabebb979ac4ba1947148040d0be1cc5/chain-storage/src/api.rs#L152);
+TODO: Should this process also include creating and spending UTXOs or should that be done during tendermint
+synchronization process?) and update `last_fetched_block` in `chain-abci`.
+
+After all the transaction data is persisted, `chain-abci` can start listening to ABCI requests block-by-block and skip
+`tx-validation` for `Transfer` and `Withdraw` transactions until `last_fetched_block`. `chain-abci` should still do a
+basic validation and update UTXOs.
```
tomtau: to reduce ambiguity how this will be "skipped" -- is something like:
* if block_height < last_fetched_block AND (tx type == `transfer` OR tx type == `withdraw`) AND tx is in storage, check output size matches AND tx inputs are unspent -> valid skip / (create tx output metadata, put in merkle tree etc.)
* if block_height < last_fetched_block AND (tx type == `transfer` OR tx type == `withdraw`) AND tx is NOT in storage -> skip as invalid
* if block_height < last_fetched_block AND tx type != `transfer` AND tx type != `withdraw`), do normal tx validation

?

tomtau (commented):
looks ok, sans some ambiguity

docs/modules/tdbe.md
tomtau: https://github.com/crypto-com/chain-docs/pull/207#discussion_r465495050

maybe just the fact they are in storage implicitly means that during "skipping", they are considered "valid"?

if abci runs, it still needs to know which tx is valid (even if it was prefetched transfer/withdraw), so that it can compute and return the same app hash for a particular block

docs/modules/tdbe.md
devashishdxt: What is "txid verification with `app_hash`"?

docs/modules/tdbe.md
```diff
@@ -35,29 +35,49 @@ and obtain old transaction data (if any)
...
-
-TODO: handling in abci -- options (probably start with option 1):
-1. record the "fetched up to" block;
-report to TM still the last processed block
-and run normal TM block-by-block syncing and skip tx-validation-enclave until the "fetched up to" block
-2. TM 0.34+ -- run state-sync for jellyfish merkle / staking state https://docs.tendermint.com/master/spec/abci/apps.html#state-sync
-and resume from the "fetched up to" block?
+### Catch-up process
+
+When a node starts, it should first check the configuration parameter `perform_catchup`. If `perform_catchup` is `true`,
```
yihuang, line 40: is this `perform_catchup` parameter necessary, can it be decided at runtime?

devashishdxt: This parameter should be provided during startup. Currently, `tdbe` assumes this to be `true` when a remote TDBE address is present in command line options.

docs/modules/tdbe.md
devashishdxt: Added

docs/modules/tdbe.md
tomtau: you get a list of valid txids --> compute the merkle tree root out of them --> is this merkle root really the component of the retrieved app hash for a particular block?

devashishdxt: Done.

tomtau (approved):

devashishdxt:
@tomtau Can we merge this?

PR #209, opened 2020-08-07
Problem (Fix #191): nodejoin tx validation is not revised with tdbe

### LINKED ISSUE #191: Problem: nodejoin tx validation is not revised with TDBE-related changes (mls init + community node rules)

yihuang:
Solution:
- add mlsinit validation rules
- add community node into staking state transition

### DIFF
--- a/docs/modules/staking-state.md
+++ b/docs/modules/staking-state.md
@@ -7,14 +7,49 @@ struct StakedState {
     bonded: Coin,
     unbonded: Coin,
     unbonded_from: Timespec,
-    validator: Option<Validator>,
+    node_details: Option<NodeDetails>,
+}
+
+enum NodeDetails {
+    Council(Validator),
+    Community(CommunityNode),
+}
+
+struct CommunityNode {
+    /// name / moniker (just for reference / human use)
+    pub name: NodeName,
+    /// optional security@... email address
+    pub security_contact: NodeSecurityContact,
+    /// leaf index in the mls tree
+    pub leaf_index: LeafSize,
 }
 
 struct Validator {
-    council_node: CouncilNode,
+    node_info: CommunityNode,
     jailed_until: Option<Timespec>,
     inactive_time: Option<Timespec>,
     used_validator_keys: Vec<(TendermintValidatorPubKey, Timespec)>,
+}
+
+impl NodeState {
+    fn validator() -> Option<&Validator> {
+        match self {
+            Council(v) => Some(v),
+            _ => None,
+        }
+    }
+    fn community() -> Option<&CommunityNode> {
+        match self {
+            Community(n) => Some(n),
+            _ => None,
+        }
+    }
+    fn node_info(&self) -> &CommunityNode {
+        match self {
+          Community(n) => n,
+          Council(v) => &v.node_info,
+        }
+    }
 }
 ```
 
@@ -22,26 +57,40 @@ struct Validator {
 
 ### Clean staking
 
-`validator.is_none()`
+`node.is_none()`
 
-### Validator
+### Community/Council node
 
-`validator.is_some()`
+`node.is_some()`
+
+One need to join the mls group to run a community or council node.
+
+#### Community node
+
+Community node is a full node but not a validator. it's able to validate transactions but not participate in block generation.
+
+`node?.community().is_some()`
+
+#### Council node
+
+`node?.validator().is_some()`
+
+> We'll use `validator` to refer to `node?.validator()?` from now on.
 
 There are several variants of it:
 
-#### Active
+##### Active
 
 `validator.inactive_time.is_none()`
 
 > **_NOTE:_** Active validator doesn't necessarily mean the final validator take effect in tendermint, please refer to
 > [Choose final validators](#choose-final-validators)
 
-#### Inactive
+##### Inactive
 
 `validator.inactive_time.is_some()`
 
-#### Jailed
+##### Jailed
 
 `validator.jailed_until.is_some()`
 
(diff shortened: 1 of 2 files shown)


docs/modules/staking-state.md
```diff
@@ -7,41 +7,90 @@ struct StakedState {
...
+    CommunityNode(NodeCommonInfo),
+}
+
+struct NodeCommonInfo {
+    /// name / moniker (just for reference / human use)
+    pub name: NodeName,
+    /// optional security@... email address
+    pub security_contact: NodeSecurityContact,
+    /// leaf index in the mls tree
+    pub leaf_index: LeafSize,
```
tomtau, line 24: In code, `NodeCommonInfo` is currently used in the node join tx payload as well -- it captured the common parts of metadata from 0.5 council node, so that the tx format isn't broken.
One thing to clarify is whether the "node join tx" payload will use a different struct (as the "leaf index" should be already contained in the Add proposal that's in "confidential init"?) for this metadata that's still compatible with the wire format or if nodejoin tx payload gets broken in 0.6 (there are potentially other breaking changes https://github.com/crypto-com/chain/pull/2071#issuecomment-669676234 )

docs/modules/staking-state.md
```diff
@@ -7,41 +7,90 @@ struct StakedState {
...
+#### Validator node
+
+`node.unwrap().validator.is_some()`
 
 There are several variants of it:
 
-#### Active
+##### Active
 
 `validator.inactive_time.is_none()`
```
tomtau, line 84: all these are then assumed to be `node.unwrap().validator`?

docs/modules/staking-state.md
```diff
@@ -51,13 +100,23 @@ There are several variants of it:
...
 - The validator pubkey/address is not already used by others, it's ok to re-use the old keys used by itself if it's a re-join from an inactive validator.
 - Not jailed if transiting from inactive validator
+- `MLSInit` is valid, refer to [transactions](#transactions.md#node-join) for detailed validation rules
+
+### From "clean staking" to "community node"
+
+#### Node join(community)
+
+The community node variant of `NodeJoinTx` transit a clean staking state to community node. the preconditions are:
+
```
tomtau, line 137: \+ not jailed?

docs/modules/transactions.md
```diff
@@ -202,17 +202,109 @@ The actual paid fee is zero.
...
-- Validator address not banned
-- Used validator address list not full when re-join with new validator key.
+
+- `staking.bonded >= minimal_required_staking or minimal_required_staking_for_community_node`
+
+- If join validator:
+
+  - Validator address not used if join validator
+
+  - Validator address not banned if join validator
```
tomtau: not banned? was that banlist of byzantine validator keys? maybe not relevant?

docs/modules/transactions.md
```diff
@@ -202,17 +202,109 @@ The actual paid fee is zero.
...
+}
+```
+
+TVE will verify the confirmation like this:
+
+```rust
+fn TVE.verify_confirmation(epoch, input_context, confirmation) {
+    // get (init_secret, commit_secret, group_context) of the epoch from TDBE,
+    // either by push or pull.
+    let (init_secret, commit_secret, group_context) = get_from_TDBE(epoch);
```
tomtau, line 290: could this be obtained just from the epoch? one will need the add proposal to compute the updated tree's hash? and will need a commit to compute the transcript_hash?

docs/modules/staking-state.md
```diff
@@ -7,41 +7,90 @@ struct StakedState {
...
+#### Community node
 
-### Validator
+Community node is a full node but not a validator. it's able to validate transactions but not participate in block generation.
 
-`validator.is_some()`
+`node.unwrap().community.is_some()`
+
+#### Validator node
```
tomtau: there's no "Validator node" -- there's "council node"

the terminology may be a bit inconsistent in this doc:
* validator == actor that may run several community nodes and council node
(also validator == node with consensus pubkey in TM validatorupdate )
* council node == application specific node metadata + TM consensus pubkey
* community node == application specific node metadata
* full node / sentry node / ... == instantiations / usage of community nodes

tomtau (changes requested):
a few comments

docs/modules/staking-state.md
yihuang: since it only transit from "clean staking" state, which means only when all the validator state has finished and cleaned up, one can join as community node.

docs/modules/staking-state.md
yihuang: yes, added a line to explain.

docs/modules/transactions.md
yihuang: can TDBE send `(init_secret, commit_secret, group_context)` of each epoch to TVE, TVE use it to verify abci request?

tomtau: there are two possibilities -- either it'll push these altogether (TVE can then derive the exporter secret), or it'll expose some `verify-confirmation-from-new-add` function

docs/modules/staking-state.md
tomtau: ok, makes sense -- so "clean staking" will be the base state that both begin with or return to.


maybe a quick state diagram would be helpful.

docs/modules/staking-state.md
yihuang: I think there's no need to bring `ConfidentialInit` into the staking state, it'll take storage space in the staking merkle tree. so it's not the same structure as the tx payload, I'll rename it to avoid confusion.
we can also figure out how to share the other parts of fields better between state and tx payload when implementing.

docs/modules/transactions.md
yihuang: removed

docs/modules/transactions.md
yihuang: added that delegating to TDBE is a potential implementation.

docs/modules/staking-state.md
yihuang: adopted the naming choice in [tdbe.md](https://github.com/crypto-com/chain-docs/blob/master/docs/modules/tdbe.md#data-type-modifications), aka: `NodeDetails` and `CommunityNode`.

docs/modules/staking-state.md
yihuang: state diagram added.

docs/modules/staking-state.md
yihuang: changed from `validator node` to `council node` in the section title, not sure how should it be changed in other places.

docs/modules/transactions.md
```diff
@@ -202,17 +202,108 @@ The actual paid fee is zero.
...
+```
+
+TVE might also delegate the request to TDBE to do the verification.
+
+After validation passed, increase the epoch, and update the tree.
+
+```rust
+state.epoch += 1;
+state.tree = updated_tree;
+```
```
tomtau: one thing to clarify is that this is only in abci, as TDBE will only update after it's been block-committed.

Normally then, this should prevent the concurrent updates (only one nodejoin will be valid -- the other one can be regenerated later against the latest epoch). Given TDBE internal state only updates after block-committed changes, it won't be possible to generate more than one valid nodejoin per block (I mean they can be generated, but only one will be valid)... so maybe one invariant is that epoch and public ratcheting tree are at most updated once per block?

tomtau (approved):

docs/modules/transactions.md
yihuang: > one thing to clarify is that this is only in abci, as TDBE will only update after it's been block-committed.
> 
> Normally then, this should prevent the concurrent updates (only one nodejoin will be valid -- the other one can be regenerated later against the latest epoch). Given TDBE internal state only updates after block-committed changes, it won't be possible to generate more than one valid nodejoin per block (I mean they can be generated, but only one will be valid)... so maybe one invariant is that epoch and public ratcheting tree are at most updated once per block?

Does the epoch validator rule forbid multiple valid node join naturally? because one can only generate new valid commit after last commit processed?

tomtau: yes, unless there's a bug to force tdbe generate payloads for not-committed epochs

yihuang: Added a validation rule that the add/commit messages are signed by left-most leaf node.

PR #217, opened 2020-08-14
Problem: possible timeout between block signing and catch-up nodejoin (fixes #215)

### LINKED ISSUE #215: Problem: possible timeout between block signing and catch-up nodejoin
"join as a community node" first, and then submit some node join to update to council node

another variant in nodejoin data as "upgrade" with just adding the validator pubkey

tomtau:
Solution: made the nodejoin process for validators two step:
1) join as a community node -- catch up with the network
2) switch to a council node

### DIFF
--- a/docs/modules/transactions.md
+++ b/docs/modules/transactions.md
@@ -202,8 +202,10 @@ The actual paid fee is zero.
 
 Action
 
-- Transit staking state from "clean staking" or "inactive(unjailed) validator" to active validator or community node (refer to  [staking
-  state](staking-state.md#node-join))
+-  Transit staking state depending on the option (refer to [staking state](staking-state.md#node-join)):
+  - from "clean staking" to community node
+  - from community node to active validator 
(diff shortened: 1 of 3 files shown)


devashishdxt (approved):

docs/modules/transactions.md
```diff
@@ -202,8 +202,8 @@ The actual paid fee is zero.
 
 Action
 
-- Transit staking state from "clean staking" or "inactive(unjailed) validator" to active validator or community node (refer to  [staking
-  state](staking-state.md#node-join))
+- Transit staking state from "clean staking" to community node (refer to  [staking
+  state](staking-state.md#node-join)) OR from community node to "inactive(unjailed) validator" or active validator
```
yihuang: maybe more clear to list out the several variants.

yihuang (approved):

lezzokafka (approved):

PR #220, opened 2020-08-18
Problem (Fix #196): details of TDBE usage of light client are missing

### LINKED ISSUE #196: Problem: details of TDBE usage of light client are missing
as mentioned in the related implementation issue:
> In particular, as the code runs inside enclaves, 1) it's trusted (-> needs to be simple and small without too many dependencies), 2) there's no "trusted" persistent storage (in particular, the host may replay the old state.

> With regards to "light client state" / trusted anchor, it implies two parts:

> 1) (no state) the full genesis.json isn't known before TDBE is compiled: https://github.com/crypto-com/chain-docs/blob/master/docs/modules/tdbe.md#genesisjson-creation -- but some parts of it should be known and should be compile-time parameters https://github.com/crypto-com/chain/issues/1996 that are cross-checked
> 2) once a node "successfully" joins (either valid genesis.json with related mls/keypackage payload OR valid committed nodejointx) OR updates (successful committed SelfUpdateProposal after NACK timeout), it can "seal" its identity+HPKE keys altogether with the "light client state" at that point

yihuang:
Solution:
- add light client verification steps

### DIFF
--- a/docs/modules/tdbe.md
+++ b/docs/modules/tdbe.md
@@ -145,39 +145,88 @@ and skipping `tx-validation`, it can potentially use Tendermint's [`StateSync`](
 for synchronizing jellyfish merkle and staking state uptil `last_fetched_block`.
 :::
 
-#### How to obtain the list of all the transactions using tendermint light client?
+## Light client
 
 To obtain all the changes in UTXO set from `last_block_height` of current node to `current_block_height` on remote node
-using tendermint light client, we use strategy similar to the one used in `client-core` for synchronizing the client.
-
-- Using `LightClient::verify_to_highest()`, retrieve `current_block_height`, `app_hash` and `block_hash` (of last block)
-  of remote node.
-- Split the block heights into a list of smaller chunks (`chunk_size` can be configurable), starting from
-  `last_block_height` of current node uptil `current_block_height` received from light client, and for each `chunk` of
-  `block_height`s:
-  - Call `block` and `block_results` on remote node's tendermint RPC for each chunk. For each `block` and
-    `block_results` in the response:
-    - Extract all the valid `Transfer` and `Withdraw` `transaction_id`s from `block_results` using code similar to
-      [this](https://github.com/crypto-com/chain/blob/8de5e08ffc6176f7ef2e726a3db1f53e783c461d/client-common/src/tendermint/types/block_results.rs#L37).
-    - Verify the root of the merkle tree of `transaction_id`s with the `app_hash` returned in `block` response. See the
-      note below for changes in `app_hash` computation.
-  - For all the `transaction_id`s in a `chunk`, establish an enclave-to-enclave mutually attested TLS stream with remote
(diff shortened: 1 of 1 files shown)


docs/modules/tdbe.md
```diff
@@ -32,16 +32,13 @@ have to follow below procedure to successfully join the network:
...
   by the node operator (see below for mode details on **Transaction signing**).
 - If the `NodeJoinTx` is accepted in a block, new node can start to operate normally (as `Commit` with Add-only proposal
   does not need to fill paths -- node need to wait for NACK timeout).
   - New node's TDBE server can apply `Welcome` message to update its state.
-  - If one wishes to join as the "council node" (have validator capabilities and responsibilities): once the local node
-  catches up fully with the latest state (i.e. chain-abci gets to the latest block), the operator can submit
-  the second `NodeJoinTx` transaction with its node's consensus public key 
 - If the `NodeJoinTx` is not accepted (this can happen when there was another `Commit` in the meantime), then restart
-  the process.
+  the proocess.
```
leejw51crypto: the process?

leejw51crypto (approved):
lgtm

docs/modules/tdbe.md
```diff
@@ -284,14 +330,14 @@ with council node and community node variants.
...
-  CommunityToCouncilNodeSwitch,
   // KeyPackage
   Genesis(Vec<u8>),
-  CommunityNodeJoin {
+  NodeJoin {
     add: Vec<u8>, // MLSPlaintext -- Add
     commit: Vec<u8> // MLSPlaintext -- Commit
 }
 ```
+TODO: extra rules for CouncilNode -> CommunityNode (needs to unbond / become inactive first?)
```
tomtau: this reverts the changes/recent revision

docs/modules/tdbe.md
```diff
@@ -32,16 +32,13 @@ have to follow below procedure to successfully join the network:
   - New node, after receiving a response from remote TDBE server, verifies (TODO: Define validations on TDBE response)
     it and starts the whole process again if the verification fails (with a different remote TDBE server). If the
     verification is successfull, it continues to the next step.
-- New node constructs `NodeJoinTx` (for community node) and submits the transaction after getting it signed
+- New node constructs `NodeJoinTx` (for council or community node) and submits the transaction after getting it signed
```
tomtau: only for community node -- then there'll be another node join after the sync up if one wishes to continue as council node

tomtau (changes requested):
looks good, but it reverted some changes from https://github.com/crypto-com/chain-docs/pull/217/files

docs/modules/tdbe.md
yihuang: yes, it's by accident, fixed.

docs/modules/tdbe.md
yihuang: it's by accident, fixed.

docs/modules/tdbe.md
yihuang: fixed

tomtau (approved):

PR #223, opened 2020-08-19
Update thaler-testnet.md

Alive29:
add systemd service


tomtau:
@Alive29 thank you for the contribution!
Currently, this tutorial page shows a warning saying:
> This page only shows the minimal setup.
> 
> You may want to (...) want to execute the processes using supervisor or equivalent etc.

so for executing the processes using systemd (or other supervisor), it'd be better to put them under the "Notes on Production Deployment" page instead (and put a link to that in "Thaler Testnet" page).

The systemd services may also need a few small improvements (~> no need for some of the environment variables; as the services are interdependent, it may not be good to individually restart them etc.)

I'll close this PR for now and I opened this issue to track this: https://github.com/crypto-com/chain-docs/issues/224 (feel free to reopen a revised PR addressing that issue though!)

PR #227, opened 2020-08-24
Problem: enclave light-client trusted anchor basis is not clear

yihuang:
Solution:
- only need to compile genesis hash into enclave, verify fetched genesis against it

### DIFF
--- a/docs/modules/tdbe.md
+++ b/docs/modules/tdbe.md
@@ -210,10 +210,21 @@ fn verify(
 
 In standard light client protocol, client use time of now to verify trusting period, since in enclave we don't have time of now, we don't verify trusting period.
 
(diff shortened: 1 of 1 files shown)


tomtau (commented):
`GENESIS_FINGERPRINT`?

devashishdxt (approved):

tomtau:
but you won't know the app_hash, as there's a cycle: in order to compute the app hash, you need the full validator state (i.e. generate initial keypackages with TDBE)

yihuang:
> but you won't know the app_hash, as there's a cycle: in order to compute the app hash, you need the full validator state (i.e. generate initial keypackages with TDBE)

yes, i changed to compile validator voting power map into enclave.

